PR #5, opened 2020-03-15
Update Gradle, Kotlin and libraries versions

magnumrocha:
- Updates:
  - Kotlin -> 1.3.70
  - Coroutines -> 1.3.4
  - Gradle -> 5.6.4
  - Firebase Android -> Updated to latest versions
  - Firebase iOS -> Updated to latest versions

- Fixed most of lint warnings


nbransby:
Hey @magnumrocha thanks for this, most of this work has taken place in master already now - you are welcome to update the PR to be compatible with the latest changes

nbransby:
Dont think we should be using [@InternalSerializationApi](https://github.com/Kotlin/kotlinx.serialization/blob/ffe216f0293231b09eea24a10aa4bc26ff6d5b90/runtime/commonMain/src/kotlinx/serialization/Annotations.kt#L103-L112), what is the goal here?

magnumrocha:
> Dont think we should be using [@InternalSerializationApi](https://github.com/Kotlin/kotlinx.serialization/blob/ffe216f0293231b09eea24a10aa4bc26ff6d5b90/runtime/commonMain/src/kotlinx/serialization/Annotations.kt#L103-L112), what is the goal here?

this is here mainly because of Kotlin Lint, without @InternalSerializationApi annotation, the Lint will be always warning about it.

nbransby:
Think you can use another annotation that doesnt propagate, where was the original function that had the warning?

magnumrocha:
@nbransby your right, I've just update the code.

nbransby:
Im going to close this as I believe most of these changes have been made on master already

PR #29, opened 2020-05-15
Firebase Auth: Add FirebaseUser.sendEmailVerification()

vanniktech:


nbransby:
just need a separate sendEmailVerification test for this

vanniktech:
Done that. Let me know whether this the way you intended it to be or not.

PR #42, opened 2020-06-18
Added firebase functions method for use of function emulator.

Reedyuk:


firebase-functions/src/jsMain/kotlin/dev/gitlive/firebase/functions/functions.kt
```diff
@@ -25,6 +25,8 @@ actual fun Firebase.functions(app: FirebaseApp, region: String) =
 actual class FirebaseFunctions internal constructor(val js: firebase.functions.Functions) {
     actual fun httpsCallable(name: String, timeout: Long?) =
         rethrow { HttpsCallableReference(js.httpsCallable(name, timeout?.let { json("timeout" to timeout.toDouble()) })) }
+
+    actual fun useFunctionsEmulator(origin: String): Unit = throw NotImplementedError()
```
nbransby, line 29: you need to implement this

Reedyuk: javascript does not expose this.

nbransby: yes it does

PR #46, opened 2020-07-01
Added starting point for github packages.

Reedyuk:
This job will automatically publish on code being merged into jvm-target.
Currently, fails silently if unable to publish, it is the responsibility for the developer to make sure their versions are published.
Will revisit when i have a better idea on failing if none are published.


build.gradle.kts
```diff
@@ -68,8 +69,8 @@ subprojects {
             name = "github"
             url = uri("https://maven.pkg.github.com/gitliveapp/packages")
             credentials {
-                username = project.findProperty("gpr.user") as String? ?: System.getenv("gpr.user")
-                password = project.findProperty("gpr.key") as String? ?: System.getenv("gpr.key")
+                username = project.findProperty("gpr.user") as String? ?: System.getenv("USERNAME")
+                password = project.findProperty("gpr.key") as String? ?: System.getenv("TOKEN")
```
Reedyuk, line 73: this is only for the private maven github packages, so belongs here and not in master.

build.gradle.kts
```diff
@@ -213,6 +224,14 @@ subprojects {
                     password = project.findProperty("sonatypePassword") as String? ?: System.getenv("sonatypePassword")
                 }
             }
+            maven {
+                name = "GitHubPackages"
+                url  = uri("https://maven.pkg.github.com/gitliveapp/packages")
+                credentials {
+                    username = project.findProperty("gpr.user") as String? ?: System.getenv("USERNAME")
```
Reedyuk, line 231: this one is for the publishing to private repo.

firebase-database/build.gradle.kts
```diff
@@ -42,6 +42,8 @@ kotlin {
                 moduleKind = "commonjs"
             }
         }
+        nodejs()
```
Reedyuk, line 45: should put this master

PR #78, opened 2020-08-21
Upgrade firebase sdk and improve iOS build process

Daeda88:
I noticed some issues with linking the project to iOS. This PR aims to resolve those issues by:

- Getting rid of downloading the entire Firebase SDK zip file from Github. Instead it uses Carthage to grab only the iOS .framework files. This is useful because:
- Upgraded to Firebase-ios-sdk 6.30, which starts splitting .framework into .xcframework
- Split linker flags in .def files to only include dependencies used by the submodules, rather than having all dependencies linked in Code. This makes it possible to just import 'Firebase/Firestore' without also downloading Auth etc.
- Upgrade Android dependencies to latest versions.


nbransby (approved):
Hey @Daeda88 this is great stuff, much appreciated!

Issue #23, opened 2020-05-11
Add FirebaseUser.updateProfile to [auth] for [ios/android]


vanniktech:
| Library               | Class                            | Member                                        | Platforms                  |
| :-----------------| :------------------------ | :----------------------------------- | :---------------------- |
| e.g auth | e.g FirebaseUser | e.g updateProfile  | Android, iOS  |

Currently, one needs to: `.android.updateProfile(UserProfileChangeRequest.Builder().setDisplayName())` which is awkward and can't be done in commonCode.

vladimirpetrovski:
Any update on this? I'm also not able to access anything related to updating the user's profile in the common module.

nbransby:
Not yet, but PRs welcome!

vanniktech:
Feel free to take a stab on this @vladimirpetrovski

Daeda88:
Im making an implementation of this. Hope to have something early next week.

PR #79, opened 2020-08-21
Increase Auth coverage

Daeda88:
Increases coverage of Firebase Auth, as requested in #23 

- Adds support for Fetching Sign in Methods
- Adds all properties to the FirebaseUser
- Adds update methods to FirebaseUser
- Adds support for Multifactor Requests
- Adds all common AuthProviders, including generic OAuth.

I'm not a Javascript developer, so I hope I didnt make any major issues there. Not al methods are tested btw, since I cant get the tests to run, and some are difficult to test anyway.

NOTE: Depends on #78


nbransby:
@Daeda88 could you merge in latest master?

Daeda88:
@nbransby Done :)

firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/auth.kt
```diff
@@ -22,71 +24,125 @@ actual class FirebaseAuth internal constructor(val android: com.google.firebase.
...
+            com.google.firebase.auth.ActionCodeResult.VERIFY_BEFORE_CHANGE_EMAIL -> Operation.VerifyBeforeChangeEmail
+            com.google.firebase.auth.ActionCodeResult.REVERT_SECOND_FACTOR_ADDITION -> Operation.RevertSecondFactorAddition
+            else -> Operation.Error
+        }
+    actual fun <T, A: ActionCodeDataType<T>> getData(type: A): T? = when (type) {
+        is ActionCodeDataType.Email -> android.info?.email
+        is ActionCodeDataType.PreviousEmail -> (android.info as? ActionCodeEmailInfo)?.previousEmail
+        is ActionCodeDataType.MultiFactor -> (android.info as? ActionCodeMultiFactorInfo)?.multiFactorInfo?.let { MultiFactorInfo(it) }
+        else -> null
+    } as? T
```
nbransby, line 98: I like the way you have implemented this! Could you take it one step further by getting rid of the type argument? you can use a reified type parameter to check what type is expected.  Also is passing the wrong type to be expected at runtime? If not (it's only a case of programmer error) the return type should not be optional instead use as! and let it throw if its the wrong type

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/auth.kt
```diff
@@ -18,42 +18,77 @@ expect fun Firebase.auth(app: FirebaseApp): FirebaseAuth
...
+
+expect class ActionCodeSettings {
+    class Builder {
+        fun setAndroidPackageName(androidPackageName: String, installIfNotAvailable: Boolean, minimumVersion: String?): Builder
+        fun setDynamicLinkDomain(dynamicLinkDomain: String): Builder
+        fun setHandleCodeInApp(canHandleCodeInApp: Boolean): Builder
+        fun setIOSBundleId(iOSBundleId: String): Builder
+        fun setUrl(url: String): Builder
+        fun build(): ActionCodeSettings
+    }
```
nbransby, line 76: To make this more kotlin-style I suggest replacing the builder pattern with a data class with optional parameters like we do with `dev.gitlive.firebase.FirebaseOptions` (this reminds me to add a section in the read me about that)

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/user.kt
```diff
@@ -0,0 +1,56 @@
...
+
+expect class UserProfileChangeRequest {
+    class Builder {
+        fun setDisplayName(displayName: String?): Builder
+        fun setPhotoURL(photoURL: String?): Builder
+        fun build(): UserProfileChangeRequest
+    }
+    val displayName: String?
+    val photoURL: String?
+}
```
nbransby, line 56: see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492703872

nbransby: actually considering this why dont we delete this class and add optional args to updateProfile of displayName and photoURL?

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/user.kt
```diff
@@ -0,0 +1,56 @@
...
+expect class UserInfo {
+    val displayName: String?
+    val email: String?
+    val phoneNumber: String?
+    val photoURL: String?
+    val providerId: String
+    val uid: String
+}
+
+expect class MetaData {
```
nbransby, line 43: rename to UserMetadata for consistency

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/user.kt
```diff
@@ -0,0 +1,56 @@
...
+    val creationTime: Long?
+    val lastSignInTime: Long?
```
nbransby, line 45: I believe we should avoid Long and use Double since long will lose its precision in JS other option would be to use https://github.com/Kotlin/kotlinx-datetime then we wouldn't face precision issues

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/multifactor.kt
```diff
@@ -0,0 +1,34 @@
...
+    val enrolledFactors: List<MultiFactorInfo>
+    suspend fun enroll(multiFactorAssertion: MultiFactorAssertion, displayName: String?)
+    suspend fun getSession(): MultiFactorSession
+    suspend fun unenroll(multiFactorInfo: MultiFactorInfo)
+    suspend fun unenroll(factorUid: String)
+}
+
+expect class MultiFactorInfo {
+    val displayName: String?
+    val enrollmentTime: Long
```
nbransby, line 17: see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492725974

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/multifactor.kt
```diff
@@ -0,0 +1,34 @@
...
+    val factorId: String
+    val uid: String
+}
+
+expect class MultiFactorAssertion {
+    val factorId: String
+}
```
nbransby, line 24: why do you think they wrap factorId in a class instead of just using factorId in enroll?

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -0,0 +1,58 @@
...
+expect object GoogleAuthProvider {
+    fun credentialWithIDAndAccessToken(idToken: String, accessToken: String): AuthCredential
+}
+
+expect class OAuthProvider constructor(provider: String, auth: FirebaseAuth = Firebase.auth) {
+    companion object {
+        fun credentialsWithAccessToken(providerId: String, accessToken: String): AuthCredential
+        fun credentialsWithIDAndAccessToken(providerId: String, idToken: String, accessToken: String): AuthCredential
+        fun credentialsWithIDRawNonceAndAccessToken(providerId: String, idToken: String, rawNonce: String, accessToken: String): AuthCredential
+        fun credentialsWithIDAndRawNonce(providerId: String, idToken: String, rawNonce: String): AuthCredential
```
nbransby, line 37: https://github.com/GitLiveApp/firebase-kotlin-sdk#named-arguments *With... methods should have their With... suffix removed as named arguments in kotlin provide the distinction between overloaded function calls (like ObjectiveC/Swift). We can keep the *With.. names in @JvmName and @JsName annotations to prevent clashes on those platforms

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -0,0 +1,58 @@
...
+expect object FacebookAuthProvider {
+    fun credentialWithAccessToken(accessToken: String): AuthCredential
+}
+
+expect object GithubAuthProvider {
+    fun credentialWithToken(token: String): AuthCredential
+}
+
+expect object GoogleAuthProvider {
+    fun credentialWithIDAndAccessToken(idToken: String, accessToken: String): AuthCredential
```
nbransby, line 29: see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492751368

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -0,0 +1,58 @@
...
+expect object EmailAuthProvider {
+    fun credentialWithEmail(email: String, password: String): AuthCredential
+}
+
+expect object FacebookAuthProvider {
+    fun credentialWithAccessToken(accessToken: String): AuthCredential
+}
+
+expect object GithubAuthProvider {
+    fun credentialWithToken(token: String): AuthCredential
```
nbransby, line 25: see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492751368

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -0,0 +1,58 @@
...
+expect class PhoneAuthCredential : AuthCredential
+
+expect class OAuthCredential : AuthCredential
+
+expect object EmailAuthProvider {
+    fun credentialWithEmail(email: String, password: String): AuthCredential
+}
+
+expect object FacebookAuthProvider {
+    fun credentialWithAccessToken(accessToken: String): AuthCredential
```
nbransby, line 21: see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492751368

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -0,0 +1,58 @@
...
+
+expect open class AuthCredential {
+    val providerId: String
+}
+expect class PhoneAuthCredential : AuthCredential
+
+expect class OAuthCredential : AuthCredential
+
+expect object EmailAuthProvider {
+    fun credentialWithEmail(email: String, password: String): AuthCredential
```
nbransby, line 17: see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492751368

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun addScope(vararg scope: String)
+    fun setCustomParameters(parameters: Map<String, String>)
+
+    suspend fun signIn(signInProvider: SignInProvider): AuthResult
+}
+
+expect class SignInProvider
+
+expect class PhoneAuthProvider constructor(auth: FirebaseAuth = Firebase.auth) {
+    fun credentialWithVerificationIdAndSmsCode(verificationId: String, smsCode: String): PhoneAuthCredential
```
nbransby, line 49: see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492751368

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -0,0 +1,58 @@
...
+    suspend fun verifyPhoneNumber(phoneNumber: String, verificationProvider: PhoneVerificationProvider): AuthCredential
+
+}
+
+expect interface PhoneVerificationProvider
+
+expect object TwitterAuthProvider {
+    fun credentialWithTokenAndSecret(token: String, secret: String): AuthCredential
```
nbransby, line 57: see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492751368

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/user.kt
```diff
@@ -0,0 +1,56 @@
...
+    val isAnonymous: Boolean
+    val isEmailVerified: Boolean
+    val metaData: MetaData?
+    val multiFactor: MultiFactor
+    val providerData: List<UserInfo>
+    val providerId: String
+    suspend fun delete()
+    suspend fun reload()
+    suspend fun getIdToken(forceRefresh: Boolean): String?
+    suspend fun linkWithCredential(credential: AuthCredential): AuthResult
```
nbransby, line 22: see see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/79/files#r492751368

nbransby (changes requested):
Thanks for submitting this and sorry for the delay in review! It's a big one!

RE the tests - we are going to need to increase the coverage to ensure stuff works in JS as I am assuming you are not using JS in your own projects? Good thing is we only need to write the tests once in commonTest and they will test all platforms (well except iOS because the tests don't actually work on iOS but thats different issue!)

Daeda88:
Thanks for the review. Ill process the remarks soon, though I'm on a different project right now so it might take one or two weeks.

thoutbeckers:
Hi @nbransby , I'm a colleague of @Daeda88 , I'm working on an upgrade to Kotlin 1.4 of this SDK (preliminary branch here: https://github.com/splendo/firebase-kotlin-sdk/tree/kotlin1.4) on top of this branch.

This brings vast improvements to coroutines for the iOS target. In this branch I have iOS tests running (some failing tests still however).

CoreyKaylor:
@thoutbeckers @nbransby is there a reason not to use the PR already submitted for kotlin 1.4? Selfishly asking of course.

thoutbeckers:
@CoreyKaylor no reason in particular (perhaps I should have made a draft PR earlier, I actually did my work before yours).

However looking at your PR, it looks like for iOS some work should be combined. Or do iOS tests actually run in your branch?

nbransby:
@thoutbeckers great news about the iOS tests! Which ones are still failing? Do you want to submit a PR for this?

@CoreyKaylor @thoutbeckers prehaps you can compare your PRs and make sure we get the best of both?

CoreyKaylor:
@thoutbeckers no I didn't make any changes to the iosTests. I would have likely submitted separate PR's, rather than putting into one big one.

rocketraman:
@thoutbeckers FYI I'm trying to build your kotlin-1.4 branch and am unable to... seems Gradle is unable to find the `kotlin.test` imports:

```
> Task :firebase-app:compileTestKotlinJs FAILED
e: /home/raman/source_ext/firebase-kotlin-sdk/firebase-app/src/commonTest/kotlin/dev/gitlive/firebase/firebase.kt: (3, 15): Unresolved reference: test
e: /home/raman/source_ext/firebase-kotlin-sdk/firebase-app/src/commonTest/kotlin/dev/gitlive/firebase/firebase.kt: (9, 6): Unresolved reference: Test
```

rocketraman:
@thoutbeckers FYI, I was able to build and package up your `kotlin-1.4` branch after erasing all the tests (for some reason the kotlin-test dependency is not being found correctly). I was then able to successfully use Firebase Auth on iOS. I'm not doing anything complex with Firebase right now except the basic auth, but that proves at least conceptually that things work! Thanks.

thoutbeckers:
Yes, there's a reason I didn't open the PR yet. We don't use `js` yet at our company so for now I just excluded JS tasks.

I'm upgrading some internal projects and other libraries to 1.4 to test all together. My colleague @Daeda88 will hopefully process feedback on this PR soon, once merged I will rebase against master and make sure JS tests work.

I'll see if I can combine changes from @CoreyKaylor(or rebase on that if it gets merged), but in particular I had sme linker related problem in 1.4.10 for native. It looks like 1.4.20 might be out soon though.

rocketraman:
@thoutbeckers Just FYI, I believe @nbransby opened a pull request on your behalf here https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/90.

nbransby:
@thoutbeckers yes to compare to the one created by @CoreyKaylor. My current preference would be to merge in @CoreyKaylor's and then we can have a look at getting the additional work you have completed on the iOS tests merged in

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
Daeda88: Since @JvmName and @JsName did not resolve clashes with relation to method signatures, I ssolved this by passing a Sealed class to the constructor instead.

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/multifactor.kt
Daeda88: Well on Android MultiFactorAssertion is an abstract class that has a subclass for PhoneMultiFactorAssertions. So I'm guessing this is future proofed for additional Multifactor solutions. However it seems to be a relatively new addition to Firebase and does not seem to fully cover all platforms yet. So I'd consider multifactor somewhat WIP for now.

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/user.kt
Daeda88: Good suggestion. Did the same for the ActionCodeSettings

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
nbransby: Yeah I got this wrong: "as named arguments in kotlin provide the distinction between overloaded function calls (like ObjectiveC/Swift)", not sure why I thought this - has this changed in 1.4 or it is still the case you can't have two methods with the same name and arg types?

Daeda88: I think it the latter. Documentation only shows examples when dealing with generics. Anyway, its solved regardless

nbransby: tbh I think:
`OAuthProvider.credentialsWithIDAndRawNonce(...)`
is a more straightforward API than:
`OAuthProvider.credentials(IdAndAccessTokenAndRawNonce(...))`
so Im going to revert this change

firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/auth.kt
nbransby: I simplified slightly by making the ActionCodeResult the sealed class

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
nbransby: In the meantime I have queried the language design decision that prevents this.. https://discuss.kotlinlang.org/t/feature-request-support-method-overloads-on-parameter-names-like-swift/19577

firebase-auth/src/iosMain/kotlin/dev/gitlive/firebase/auth/auth.kt
```diff
@@ -106,19 +167,47 @@ private fun <T, R> T.throwError(block: T.(errorPointer: CPointer<ObjCObjectVar<N
...
+        } else if(error != null) {
+            job.completeExceptionally(error.toException())
+        } else {
+            job.complete(default)
+        }
+    }
+    return job.await()
+}
+
+internal suspend fun <T, R> T.awaitExpectedResult(function: T.(callback: (R?, NSError?) -> Unit) -> Unit): R {
```
nbransby, line 196: Which API call did you need to make this change for?

Daeda88: Anything that deals with optional. The original code would not complete if the completion block returned an optional result. I ran into a few instances where the callback from the iOS SDK returned both value and error as null, which is by design since null is a valid response type. In that case, the original awaitResult would never complete and cause the coroutine calling the sdk to suspend indefinitely.

Now the distinction between `awaitResult(function)` `awaitResult(default, function)` and `awaitExpectedResult(function)` jut handless optionals differently. I guess `awaitResult(default, function)` could be removed in place of just doing a `?:` assignment afterwards. The awaitExpectedResult is there for API calls where the callback receives either `(value, null)` or `(null, error)`, but never `(null, null)`. Its essentially a force unwrap. If for some reason the api does return `(null, null)` it should finish with an error.

nbransby: Thanks, which iOS API call is it that can return a null result specifically?

Daeda88: I dont fully recall, but looking at my original code:
`Auth.fetchSignInMethodsForEmail`
`User.getIdToken`
`User.unlink`

Essentially anywhere I called `awaitResult` the return type can be optional and any place I called `awaitExpectedResult` does not expect a null type (hence the Expected result)

nbransby: Can't see in the docs of those calls that they would never null in successful cases also if they did then you make the return type nullable and return the null to the client

Daeda88: Well they are, I ran into it when testing or I would not have changed things.

And you can't do the latter because that makes infering the type impossible, since the callback type will always be of the form `(R?, NSError?)`, while the return type could be either R or R?, depending on whether you want the optional or non-optional result. Hence we have:
`internal suspend fun <T, R> T.awaitResult(function: T.(callback: (R?, NSError?) -> Unit) -> Unit): R?`
and
internal suspend fun <T, R> T.awaitExpectedResult(function: T.(callback: (R?, NSError?) -> Unit) -> Unit): R

Notice how one returns R? and the other R.

In the original method it was `internal suspend fun <T, R> T.awaitResult(function: T.(callback: (R?, NSError?) -> Unit) -> Unit): R`, returning R as a non-optional.

And then on the fork you made of this branch it changed to `
internal suspend fun <T, R> T.awaitResult(function: T.(callback: (R, NSError?) -> Unit) -> Unit): R` where R will always be optional since the callback is always optional. This actually results in the iOS code not compiling at the moment, because the actual methods expect a non-optional return type but they receive an optional one.

So I would suggest you keep this change, because I dont think force unwrapping the result is a better solution.

Daeda88: As for the documentation, Android/Kotlin is a bit clearer wrt optionality. Here's the result for fetchSignInMethods:
https://firebase.google.com/docs/reference/kotlin/com/google/firebase/auth/SignInMethodQueryResult.html

`abstract MutableList<String!>? | getSignInMethods()`

nbransby: I commited to that branch to switch from windows to mac to finish this off as it was too hard guessing the correct calls 😆 

Anyway finished it now - you can still have a single awaitResult function for both R and R? by [casting the return type](https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/92/commits/74c13fce3c5ec4ad53f39a3aa91156c6b74e27c1#diff-d29e748eb38c42986c332fcbf77831123ae810cf836b06e59de2ee42e2acddb0R137)

PR #92, opened 2020-10-20
Increase auth coverage

nbransby:


Daeda88:
@nbransby Thanks for the feedback/changes.

I also made a small additional PR that fixes some coroutine cancellation exceptions I kept having due to concurrency issues: https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/93

I recommend merging that in before merging

PR #100, opened 2020-11-02
prevent accidental removal of photo URL

tynn:
This PR fixes two issues with `updateProfile()` at once

 1. defaulting to null for display name and photo url will unset the value when set
 2. `photoUrl` was not used on _Android_


nbransby (approved):

nbransby:
Thanks for this!

Issue #124, opened 2020-12-23
Add FieldPath.documentId() to firestore for every platform


leoxs22:
| Library               | Class                            | Member                                        | Platforms                  |
| :-----------------| :------------------------ | :----------------------------------- | :---------------------- |
| firestore |  FieldPath | documentId  | ios, android, js |

I'm making the android version, but I'm having problems with cocoapods, so I can't know if it compiles. Can someone else do the other 2?

Thanks!

tiagonuneslx:
@nbransby Why is this issue closed? It's still not possible to do FieldPath.documentId(), right?

I tried doing:
```
firestore.collection("things")
    .where(FieldPath.documentId, inArray = documentIds)
```

Doesn't recognize any FieldPath object...

Then I tried:
```
firestore.collection("things")
    .where(FieldPath().documentId, inArray = documentIds)
```

Fails in runtime: `java.lang.IllegalArgumentException: Invalid field path. Provided path must not be empty`

tgyhlsb:
As of today, this issue remains and should not be closed.
The author of the only PR linked here mentioned that they only did the Android part.

tgyhlsb:
By inspecting the firebase iOS repo, I've found that using `"__name__"` works:
https://github.com/firebase/firebase-ios-sdk/blob/973f8dfdda5d51f306771dcade74172bd5ee7d43/Firestore/core/src/model/field_path.h#L46


```kotlin
firestore.collection("things")
    .where(FieldPath("__name__") inArray = documentIds)
```

PR #125, opened 2020-12-23
Support for documentId in FieldPath (Firestore)

leoxs22:
#124 

I've made the android part, but I don't know how to do the other 2 :S


mjmarathon:
How is one intended to use this?

tiagonuneslx:
I don't get it either... (#124)

PR #135, opened 2021-01-07
bumped version and changed to umd

Reedyuk:


mihbor:
Hi, is this going to be merged to master and released?
Trying to use signInWithPopup from a UMD js module like so:
`Firebase.auth.js.signInWithPopup(firebase.auth.GoogleAuthProvider())`
is giving me "When accessing module declarations from UMD, they must be marked by both @JsModule and @JsNonModule" error.
I think this change would fix that?

PR #142, opened 2021-01-28
added AuthTokenResult

### LINKED ISSUE #107: Add FirebaseUser.idToken with claims to firebase-auth
| Library               | Class                            | Member                                        | Platforms                  |
| :-----------------| :------------------------ | :----------------------------------- | :---------------------- |
| user | FirebaseAuth | idToken  | Android, iOS  |

The current version of `idToken` returns only the token itself. The underlying library does provide access to the claims present in the token via the claims property on both iOS and Android. It would be nice if there was a way to get those claims without having to parse the token again.

Reference documentation for APIs in question:

Android: https://firebase.google.com/docs/reference/android/com/google/firebase/auth/GetTokenResult#getClaims()

iOS: https://firebase.google.com/docs/reference/swift/firebaseauth/api/reference/Classes/AuthTokenResult#claims

Note that both platforms provide some additional convenience properties as well, such as the auth timestamp, expiration timestamp, issued at timestamp, and sign-in provider. Ideally, all of those would be available, but from my perspective the claims are the most important and any other attribute should be obtainable from the claims.

suntrix:
Fixes #107

### DIFF
--- a/firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/auth.kt
+++ b/firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/auth.kt
@@ -42,9 +42,9 @@ expect class AuthResult {
 }
 
 expect class AuthTokenResult {
-    //    val authTimestamp: Long
+//    val authTimestamp: Long
     val claims: Map<String, Any>
-    //    val expirationTimestamp: Long
+//    val expirationTimestamp: Long
 //    val issuedAtTimestamp: Long
     val signInProvider: String?
     val token: String?
--- a/firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/auth.kt
+++ b/firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/auth.kt
@@ -41,6 +41,15 @@ expect class AuthResult {
     val user: FirebaseUser?
 }
 
+expect class AuthTokenResult {
+    //    val authTimestamp: Long
+    val claims: Map<String, Any>
+    //    val expirationTimestamp: Long
+//    val issuedAtTimestamp: Long
+    val signInProvider: String?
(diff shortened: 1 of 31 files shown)


firebase-common/src/jsMain/kotlin/dev/gitlive/firebase/externals.kt
```diff
@@ -68,6 +68,17 @@ external object firebase {
             fun onAuthStateChanged(nextOrObserver: (user.User?) -> Unit): () -> Unit
             fun onIdTokenChanged(nextOrObserver: (user.User?) -> Unit): () -> Unit
         }
+
+        abstract class IdTokenResult {
+            val authTime: String
+            val claims: Map<String, Any>
```
nbransby, line 74: this is not a map it will be Json

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/auth.kt
```diff
@@ -41,6 +41,15 @@ expect class AuthResult {
     val user: FirebaseUser?
 }
 
+expect class AuthTokenResult {
+    //    val authTimestamp: Long
+    val claims: Map<String, Any>
+    //    val expirationTimestamp: Long
+//    val issuedAtTimestamp: Long
+    val signInProvider: String?
+    val token: String?
```
nbransby, line 50: can they be null?

nbransby (changes requested):

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/auth.kt
suntrix: Unfortunately the properties in Android SDK are marked as `@Nullable`, so didn't want to risk a crash.

Issue #141, opened 2021-01-22
Add FirebaseApp.initialize to firebase.common for all platforms


Lamartio:
Hey all!,

Thanks for creating this amazing wrapper! I run into an issue though: When I want to login (on web), I get the error: `Error: Be sure to include authDomain when calling firebase.initializeApp(), by following the instructions in the Firebase console.`. This is the snippet that caused the error:
```
Firebase
    .auth(app)
    .js
    .signInWithPopup(firebase.auth.GoogleAuthProvider())
    .await()
```

When I initialize the app I am missing the option `authDomain` in the `dev.gitlive.firebase.FirebaseOptions`. Could you at this property to the object? 

| Library               | Class                            | Member                                        | Platforms                  |
| :-----------------| :------------------------ | :----------------------------------- | :---------------------- |
| common | FirebaseApp | initializeApp | all |

mihbor:
Encountered this issue when trying to do the same.
I suggest renaming this issue to "add authDomain to FirebaseOptions" for easier discoverability.

mihbor:
I made a stab at a PR to add authDomain to FirebaseOptions (above). I haven't managed to get master to build locally I'm afraid, so I wasn't able to test this either.

PR #149, opened 2021-02-14
#141 add authDomain to Firebase options

mihbor:


Daeda88 (changes requested):
Missing implementation for Android & iOS

mihbor:
@Daeda88 do the Android and iOS APIs take authDomain as a parameter? I couldn't find it in their docs but if I missed it please point me in the right direction and I'll update the PR.

Daeda88:
Ehr, you're absolutely right.

Daeda88 (approved):

PR #153, opened 2021-02-17
Where for DocumentReference

avdyushin:
So I can use filter with FirestoreReference:

```
query.where(
    "my_field",
    Firebase.firestore.collection("my_collection").document("my_document_id")
)
```


avdyushin:
Looks like JS tests are failed in `master` branch

PR #170, opened 2021-04-13
Fix iOS Firebase add not working

Daeda88:
Calling Add on Firestore fails to complete due to missing callback


Daeda88:
Not sure why the ci is failing but since it's on the js side that shouldn't be related to this pr

avdyushin (approved):

PR #198, opened 2021-07-05
Fix iOS tests, upgrade iOS Firebase SDK

shepeliev:
After updating Firebase iOS SDK some tests fail because `ld` could not find the required frameworks. 
Another problem, FirebaseFirestore 8.1.0 has Carthage [bug](https://github.com/firebase/firebase-ios-sdk/issues/8112) that is the cause of missing gRPC-C++ framework. This PR fixes these issues. Also, I've enabled iOS testing at GitHub actions. Unfortunately, auth tests fail on iOS simulator, therefore I've added a skip flag for ignoring tests on iOS.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -30,12 +30,20 @@ jobs:
       run: ./gradlew assemble
     - name: Run JS Tests
       run: ./gradlew cleanTest jsTest
-    - name: Upload test artifact
+    - name: Upload JS test artifact
       uses: actions/upload-artifact@v2
       if: failure()
       with:
-        name: "JSTest Report HTML"
+        name: "JS Test Report HTML"
         path: "firebase-firestore/build/reports/tests/jsTest/"
+    - name: Run iOS Tests
+      run: ./gradlew cleanTest iosTest
+    - name: Upload iOS test artifact
+      uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        name: "iOS Test Report HTML"
+        path: "firebase-firestore/build/reports/tests/iosTest/"
     - name: Run Android Instrumented Tests
       uses: reactivecircus/android-emulator-runner@v2
       with:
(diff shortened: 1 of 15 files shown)


Reedyuk (approved):

nbransby:
cant we skip the ios tests at the build script instead of at the test level @shepeliev @Reedyuk ?

Reedyuk:
Yeah we could add something to skip ios tests in the build gradle if you want this?

nbransby:
@Reedyuk yeah as that will a lot easier to maintain that at code level, but its just the auth tests we want to skip correct?

nbransby:
@shepeliev last time I checked I thought the ios test didn't work at all - they just froze, in fact the `runTest` implementation was just a guess:

```
actual fun runTest(test: suspend () -> Unit) = runBlocking {
    val testRun = MainScope().async { test() }
    while (testRun.isActive) {
        NSRunLoop.mainRunLoop.runMode(
            NSDefaultRunLoopMode,
            beforeDate = NSDate.create(timeInterval = 1.0, sinceDate = NSDate())
        )
        yield()
    }
    testRun.await()
}
```

Are you saying this actually works now?

shepeliev:
> @shepeliev last time I checked I thought the ios test didn't work at all - they just froze, in fact the `runTest` implementation was just a guess:
> 
> ```
> actual fun runTest(test: suspend () -> Unit) = runBlocking {
>     val testRun = MainScope().async { test() }
>     while (testRun.isActive) {
>         NSRunLoop.mainRunLoop.runMode(
>             NSDefaultRunLoopMode,
>             beforeDate = NSDate.create(timeInterval = 1.0, sinceDate = NSDate())
>         )
>         yield()
>     }
>     testRun.await()
> }
> ```
> 
> Are you saying this actually works now?

Looks like it was an accurate guess ) because it works for me. Actually, I'm a newbie in KMP and Android background, so I can't say if iOS `runTest` implementation is propper, but I use it in another my project and as I said it works well enough.

PR #199, opened 2021-07-08
Database transactions

michaelprichardson:


Reedyuk (approved):

firebase-database/src/androidMain/kotlin/dev/gitlive/firebase/database/database.kt
```diff
@@ -188,8 +189,30 @@ actual class DatabaseReference internal constructor(
...
+
+            override fun onComplete(
+                error: DatabaseError?,
+                committed: Boolean,
+                snapshot: com.google.firebase.database.DataSnapshot?
+            ) {
+                if (error == null && snapshot != null) {
+                    deferred.complete(Result.success(DataSnapshot(snapshot)))
+                } else {
+                    deferred.complete(Result.failure(Throwable(error?.message)))
```
nbransby, line 208: we've got a database specific exception class you can throw here

nbransby: I dont think you need to use the Result type and use https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/complete-exceptionally.html instead of throwing

firebase-database/src/androidMain/kotlin/dev/gitlive/firebase/database/database.kt
```diff
@@ -188,8 +187,30 @@ actual class DatabaseReference internal constructor(
...
         .run { if(persistenceEnabled) await() else awaitWhileOnline() }
         .run { Unit }
-}
 
+    actual suspend fun <T> runTransaction(strategy: KSerializer<T>, transactionUpdate: (currentData: T) -> T): DataSnapshot {
+        val deferred = CompletableDeferred<DataSnapshot>()
+        android.runTransaction(object : Transaction.Handler {
+
+            override fun doTransaction(currentData: MutableData) =
+                Transaction.success(transactionUpdate(decode(strategy, currentData)) as MutableData)
```
pauminku, line 196:
```suggestion
           override fun doTransaction(currentData: MutableData): Transaction.Result {
                currentData.value = currentData.value?.let {
                    transactionUpdate(decode(strategy, it))
                }
                return Transaction.success(currentData)
            }
```

I'm suggesting this change because I think transactions are not working. 
We need to work with `currentData.value`, not with `currentData` itself.
Also we need to tolerate `null` value as described in the [Firebase documentation](https://firebase.google.com/docs/database/android/read-and-write#save_data_as_transactions).
I'm happy to help to update and merge this PR but I'm not sure I'll be able to do the same in js and iOS.

pauminku:
Should this PR work on iOS? I'm getting this error: `Uncaught Kotlin exception: kotlin.native.IncorrectDereferenceException: illegal attempt to access non-shared <object>@xxxx from other thread`

Reedyuk:
> I think to make this work on iOS we'll need to freeze the callback blocks: https://youtrack.jetbrains.com/issue/KT-43566

I wouldnt freeze it, you should be able to enable the new memory model and disable freezing.

https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md

```
kotlin.native.binary.memoryModel=experimental
kotlin.native.binary.freezing=disabled
```

PR #205, opened 2021-08-02
Implementation/remote config

corrado4eyes:


suntrix:
@corrado4eyes  You might wanna update your branch with the latest master, RemoteConfig is already implemented there.

corrado4eyes:
@suntrix You're right, didn't noticed that I will close this PR.

Issue #38, opened 2020-06-12
Open the library for additional serialization approaches


brezinajn:
It would be a good idea to at least expose additional API to be able to plug other serialization providers.
Now we have
```Kotlin
fun <T> WriteBatch.set(documentRef: DocumentReference, strategy: SerializationStrategy<T>, data: T, merge: Boolean): WriteBatch
fun <T> DocumentSnapshot.data(strategy: DeserializationStrategy<T>): T
```
Proposed API
```Kotlin
inline fun <T> WriteBatch.set(documentRef: DocumentReference, crossinline encode: (T)->Any?, data: T, merge: Boolean): WriteBatch
inline fun <T> DocumentSnapshot.data(crossinline parse: (Any?)->T): T
```
By providing this API we should be able to plug in any serialization library (including `kotlinx.Serialization`).

teamhubuser:
Hi @brezinajn what serialization provider are you using?

brezinajn:
When it's my decision kotlinx.Serialization. But in some older projects we have both Moshi and Jackson. We even have some projects using Helios by 47degrees. The thing is just to provide ability to plug serialization solution that suits the needs of your project.

nbransby:
Problem with such a change is what format is the data passed to the parse function in? If we pass the raw format from underlying Firebase library it will be platform specific

brezinajn:
> Problem with such a change is what format is the data passed to the parse function in? If we pass the raw format from underlying Firebase library it will be platform specific

That is a good point. It should be possible in the same manner as with current serializer/deserializer implementation.

```Kotlin
fun <T> dev.gitlive.firebase.decode(strategy: DeserializationStrategy<T>, value: Any?): T {
    require(value != null || strategy.descriptor.isNullable) { "Value was null for non-nullable type ${strategy.descriptor.serialName}" }
    return FirebaseDecoder(value).decode(strategy)
}
```
What this function does is, it takes (multiplatform) deserializer and some value that came from firestore and runs the value through deserializer. There should be no need to force a specific deserializer.

```Kotlin
fun <T> decode(parse: (Any?)->T, value: Any?): T = parse(value)

fun <T> decode(strategy: DeserializationStrategy<T>, value: Any?): T = parse({ data ->
    require(value != null || strategy.descriptor.isNullable) { "Value was null for non-nullable type ${strategy.descriptor.serialName}" }
    FirebaseDecoder(data).decode(strategy)
}, value)
```
[...]

PR #214, opened 2021-08-28
added DocumentSnapshot.dataMap() (fixes #186)

### LINKED ISSUE #186: Add DocumentSnapshot.entries to Firestore
With the current API, there is no way to retrieve unknown fields from a `DocumentSnapshot`, only known fields can be retrieved (e.g. with `contains()` and `get()`).

The `DocumentSnapshot` needs to also expose a map-like API, like below:

```Kotlin
val size: Int
val entries: Set<Pair<String, Any>>
val keys: Set<String>
val values: Collection<Any>
```

This allows enumeration of all fields, not just those known at compile-time.

suntrix:


suntrix:
@nbransby / @Reedyuk - Can you review?

firebase-firestore/src/androidMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -413,6 +413,8 @@ actual class DocumentSnapshot(val android: com.google.firebase.firestore.Documen
 
     actual fun <T> data(strategy: DeserializationStrategy<T>) = decode(strategy, android.data)
 
+    actual fun dataMap(): Map<String, Any?> = android.data ?: mapOf()
```
Reedyuk, line 416: minor - this could be just emptyMap()

firebase-firestore/src/iosMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -381,6 +381,8 @@ actual class DocumentSnapshot(val ios: FIRDocumentSnapshot) {
 
     actual fun <T> data(strategy: DeserializationStrategy<T>) = decode(strategy, ios.data())
 
+    actual fun dataMap(): Map<String, Any?> = ios.data()?.map { it.key.toString() to it.value }?.toMap() ?: mapOf()
```
Reedyuk, line 384: emptyMap

Reedyuk (approved):

firebase-firestore/src/androidMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
suntrix: I totally missed it!

nbransby:
This one slipped past me but I believe dataMap() is flawed due to the reason stated here https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/38#issuecomment-644874141 and should probably be removed to avoid exposing platform specific types to the multiplaform code.  Forcing usage of kotlinx-serialization is the only way to ensure the api remains platform dependent

PR #252, opened 2021-11-03
updated Firebase SDKs

suntrix:


suntrix:
@nbransby / @Reedyuk can you take a look?

firebase-config/src/jsTest/kotlin/dev/gitlive/firebase/remoteconfig/FirebaseRemoteConfig.kt
```diff
@@ -9,8 +9,8 @@ actual fun runTest(test: suspend () -> Unit) = GlobalScope
     .promise {
         try {
             test()
-        } catch (e: dynamic) {
-            (e as? Throwable)?.log()
```
nbransby, line 13: Throwable maps to Error js so this won't catch other non-errors that might be thrown (I observed the firebase libs throwing non-errors)

suntrix: I'll revert that, thx!

suntrix: @nbransby done

nbransby:
as you can see from this PR https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/224/files we actually removed that ktx dependency intentionally.

my preference would be to remove all the ktx dependencies - it pulls in extra dependencies aka extra bloat for the end user of this library which they don't gain from in anyway and we don't really need to use the ktx extensions internally in the lib

suntrix:
@nbransby I removed all -ktx dependencies & rebased on top of master

suntrix:
@nbransby / @Reedyuk can you take a look?

nbransby (approved):

PR #254, opened 2021-11-16
updated Firebase JS SDKs

suntrix:


suntrix:
@nbransby / @Reedyuk can you take a look?

nbransby:
Thanks @suntrix ! What's the compat prefix for?

suntrix:
Firebase JS SDK v9 is not compatible with v8, but they added compatibility layer so we can update to v9 without changing the implementation & then gradually update the implementations (https://firebase.google.com/docs/web/modular-upgrade).

nbransby (approved):

PR #244, opened 2021-10-19
fix compilation with xcode 13

muwasi:
Current repo doesn't compile with Xcode 13. This fix it

PR #256, opened 2021-11-19
Workaround for XCode 13 Fixes#255

litclimbing:
This allows the cinterop tool to run correctly with XCode 13. See https://youtrack.jetbrains.com/issue/KT-48807

I believe this is fixed with Kotlin 1.6.0 but this library is using 1.5.31 for now.


nbransby:
There seems to be a lot more than fixing xcode 13 in here?

Also isnt this already fixed in https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/244 ?

nbransby:
Close in favour of #245

PR #257, opened 2021-11-23
Added firebase-installations

suntrix:


nbransby:
@suntrix you are contributor #1! 🥇 

Isn't installations just  part of firebase-app? Would it not be better to add the API there instead of a separate module?

suntrix:
> @suntrix you are contributor #1! 🥇
> 
> Isn't installations just part of firebase-app? Would it not be better to add the API there instead of a separate module?

🥳 

I initially did that, but then I noticed they have a separate namespace for it & moved it to separate module.
TBH I'm good with either, I can move it back if you'd prefer that.

firebase-installations/build.gradle.kts
```diff
@@ -0,0 +1,162 @@
...
+        val commonMain by getting {
+            dependencies {
+                api(project(":firebase-app"))
+                implementation(project(":firebase-common"))
+            }
+        }
+
+        val androidMain by getting {
+            dependencies {
+                api("com.google.firebase:firebase-installations")
```
suntrix, line 133: @nbransby Actually there's a separate package for Android, so probably makes sense to leave it as separate module.

suntrix: @nbransby Can you take a look? I have some more PRs in my pipeline :)

suntrix:
@nbransby any possibility to merge it this week?

nbransby (approved):

PR #259, opened 2021-12-09
Added "parent" field to document and collection references

litclimbing:
Added the parent field to allow easy traversal of the references. I don't fully understand the JS side so if someone wants to check that out that would be good.


nbransby:
Hi @litclimbing looks good, just need a simple test - the tests are also multiplatform and are written in commonTest which means you can write the test in kolin and the CI will run it on all platforms for you so we know it works on JS too

nbransby (approved):

PR #260, opened 2021-12-09
Platform/jvm

LandryNorris:
Add support for the Admin SDK on a pure JVM platform. The Firestore and Realtime Database modules map very well between the client and admin SDK, so I have added them in this PR. The other modules do not map as well, so a different solution may be better to handle those. I attempted to minimize changes to the common code as much as possible, but had to do a minor refactor on the FirebaseOptions definition to support both mobile and admin. Please let me know if any changes should be made or if this would be better as a separate library.


LandryNorris:
As of now, the Admin SDK has less support for emulators than the client SDKs, so I have set it up using a real Firebase Project. The credentials would have to be replaced in order to set up tests with a different Firebase Project

nbransby:
Hi @LandryNorris thanks for this PR! After looking through the code I believe it would be best to put this into a separate module, we could use `dev.gitlive.firebase-admin` 

Reasoning for this is:
- A major goal for this library is to provide a single API for firebase that can be consumed directly from common code in a kotlin multiplatform project. A single project is likely to either be clientside (mobile/desktop/web app) or serverside not both.
- If you expanded the support to the node firebase admin sdk, you would have a clash between jt targets
- Adding it to the existing modules is a breaking change and extra api noise for client library consumers

I see value in adding a seperate firebase-admin module with targets for jvm and node as serverside kotlin depending on firebase-admin could more easily shared by java and node environments (although somewhat less valuable than client side multiplatform)

LandryNorris:
I can see the lack of value in sharing the code on the server side. The primary thing I'm looking for is the serialization and coroutine integrations y'all have added, which makes coding for firebase faster. I can switch the design to use modules for admin (perhaps a firebase-admin, firebase-admin-database, firebase-admin-firestore or something similar?) unless making it a separate project would be a better way to do that, leaving this codebase for client side.

manosbatsis:
@LandryNorris Can this PR be used for jvm targets in a multiplatform project? We're after supporting android, iOS web and (JVM) desktop variants, this looks like it can help?

LandryNorris:
In theory, yes, but it uses the admin SDK under the hood, since there's no non-Android client SDK for jvm as far as I can tell, so I personally wouldn't use it for a desktop app available to the public, but if that's OK or if this app is for internal use, desktop jvm should work just fine.

nbransby:
Actually it uses a pure JVM port of the Android SDK which we haven't open sourced yet as it's got some rough edges

tieskedh:
is there any info on this?

PR #277, opened 2022-03-26
updated Kotlin to 1.6.10, Android target SDK to 31, Android Gradle Plugin to 7.1.0 & dependencies

suntrix:


suntrix:
@nbransby The build seems to be stuck, might wanna re-run this.

suntrix:
@nbransby / @Reedyuk  - Can you guys review/merge this? I have another PR that will add support for other Apple platforms :)

suntrix:
@nbransby 👋

suntrix:
Done! 
Hopefully you won't freak out when you'll see the next PR ;)

rocketraman:
Compiling this version of the code results in:

```
> Task :firebase-auth:compileTestDevelopmentExecutableKotlinJsIr
e: Module "dev.gitlive:firebase-auth" has a reference to symbol dev.gitlive.firebase.auth/FirebaseUser.displayName.<get-displayName>|-7122534302014937473[1]. Neither the module itself nor its dependencies contain such declaration.
```

which seems to caused by https://youtrack.jetbrains.com/issue/KT-48836.

Changing in `FirebaseUser`:

```
suspend fun updateProfile(displayName: String? = this.displayName, photoUrl: String? = this.photoURL)
```

to

```
suspend fun updateProfile(displayName: String?, photoUrl: String?)
```

works around the issue.

suntrix:
@nbransby / @rocketraman - I can repro the compilation error & I applied the fix.

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/user.kt
```diff
@@ -28,7 +28,7 @@ expect class FirebaseUser {
     suspend fun updateEmail(email: String)
     suspend fun updatePassword(password: String)
     suspend fun updatePhoneNumber(credential: PhoneAuthCredential)
-    suspend fun updateProfile(displayName: String? = this.displayName, photoUrl: String? = this.photoURL)
+    suspend fun updateProfile(displayName: String? = null, photoUrl: String? = null)
```
rocketraman, line 31: @suntrix I'd recommend not providing defaults here at all -- I'm assuming that if null is passed here it will remove any existing values, therefore changing the signature this way can cause data loss if the caller isn't aware of the change. Better for callers to upgrade and fail fast with a compile error... after which some docs/explanation would be handy for them to understand why that is now failing to compile.

nbransby: This was a good spot, any reason for changing this from = this.x to = null ?

suntrix: @nbransby I just saw you guys already did that in other places in the same file.
@rocketraman Yeah, just checked inside the Android SDK and your assumption was right, so it's safe to assume it can affect all platforms (PR updated).

nbransby: Where are the other places? Do we need to change these lines at all?

rocketraman: @nbransby The underlying reason for this change is explained in my comment above: https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/277#issuecomment-1092269329.

suntrix: @nbransby Sorry, for the poor explanation there. I meant that when changing that function signature to `= null` I saw some other functions in this file that had nullable params & defaulted to null, so I blindly followed that, which was wrong (thx @rocketraman for catching that!).

About the second question, short answer - yes we need to change them to what @rocketraman suggested. Probably a good idea making that KotlinJS IR compilation a part of the PR verification as well. Let me know if you're good with me adding this to github actions.

nbransby: > Let me know if you're good with me adding this to github actions.

Yes please 👍

nbransby: So we shouldn't make a breaking change to the public api due to a bug in the IR compiler, in fact these bugs were the reasons I was reluctant to add IR compiler support to the project, we are not using IR in production and neither are JetBrains (Space still on legacy compiler)

Best workaround would be to add a default sentinel value and check for that in the implementations of the function as long as that doesn't break the IR compiler also.

So something like
```
val UNCHANGED = ""

suspend fun updateProfile(displayName: String? = UNCHANGED photoUrl: String? = UNCHANGED)

actual fun updateProfile(displayName: String?, photoUrl: String?) {
{
  ...if(displayName !== UNCHANGED) ...
}
```

suntrix: @nbransby I'll revert the changes in this PR & I'll create a separate one to address this issue, since my main goal here was Kotlin upgrade not to fix IR compiler bugs.

rocketraman: > in fact these bugs were the reasons I was reluctant to add IR compiler support to the project, we are not using IR in production and neither are JetBrains (Space still on legacy compiler)

I for one am glad you did anyway @nbransby. The IR compiler is pretty clearly the future and while you may not use IR internally as of yet, unlike Space, this project is a *library*. It is important that libraries support the IR compiler because only that will encourage use of the new compiler in the real world, and only that in turn will ensure the new compiler becomes as stable as possible when it is finally released i.e. bugs like this one will get discovered and get more visibility. So thanks!

suntrix:
@nbransby everything's ready

suntrix:
@nbransby 👋

nbransby:
jsut waiting on #286 so we can do one more 1.5 based release before upgrading to 1.6

firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/user.kt
rocketraman: > @nbransby I'll revert the changes in this PR & I'll create a separate one to address this issue

@suntrix Did you ever create a new PR to address this issue?

The problem I described in https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/277#issuecomment-1092269329 still exists in 1.6.2, and the release of 1.6 would have been a perfect time to simply remove the default parameters for this method, as very clearly they should be, given that null removes data -- the only alternative is the sentinel value proposed earlier by @nbransby but that seems like a real hack.

Furthermore, the other `update` methods also do not have default parameters, so this would make the API consistent: https://github.com/GitLiveApp/firebase-kotlin-sdk/blob/master/firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/user.kt#L29-L30.

This would fix https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/296.

nbransby: Have they not fixed the underlying bug in the IR compiler yet?

rocketraman: Though I also understand leaving it as-is if you don't care about the IR support...

rocketraman: > Have they not fixed the underlying bug in the IR compiler yet?

No they haven't :(

nbransby:
Is there even a issue for it on youtrack?

rocketraman:
> Is there even a issue for it on youtrack?

https://youtrack.jetbrains.com/issue/KT-48836

PR #293, opened 2022-04-20
More native targets

suntrix:


suntrix:
@nbransby Can you take a look at this and let me know if you'd want any changes?

I had to move `iosMain/iosTest -> nativeMain/nativeTest` which will cause potential merge conflicts with any new PRs.

I tried to divide the whole PR into logical chunks with the main stuff contained in the first 2 commits).

suntrix:
@nbransby 👋

suntrix:
@nbransby Any chance for review / merge of this PR?

nbransby:
Hi @suntrix sorry for the delay in reviewing and thank you for another awesome pr! I have a question on the architecture here - should we consider just a single target named apple since this is the terminology firebase use (https://firebase.google.com/docs/ios/setup)? I think we should replace what you have as the 'native' target as 'apple' because maybe one day someone will add support for the unity or C++ firebase sdks which would also be native target but wouldn't bind to the apple (iOS+) sdk

What benefit do the additional faux targets bring? Specially ios, tvos, macos (I noticed the apple sdk also supports watchos, should we add that?). Are there some kind of default target set for Kotlin/Native that predefine these?

suntrix:
No worries.

I was following the hierarchical project structure with source sets & naming (example in the docs which is based on kotlinx.coroutines: https://kotlinlang.org/docs/multiplatform-share-on-platforms.html#share-code-in-libraries).
The main benefit with following this is that the lib can be used by anyone that for example would use Kotlin multiplatform for building just macOS/tvOS apps & they can still include the dependencies in `commonMain`.

We can't use a single target, because some of the Firebase SDKs don't expose all APIs to all Apple platforms, that's why you can see in firebase-auth module I had to add `iosMain` for any iOS-specific APIs & remove them from `nativeMain`. I also had some issues with watchOS that's also why it's not in this PR, but it will happen sooner or later :)

I'm also looking at adding the C++ SDKs for linux, but these technically will allow you to support:
* Android -> https://firebase.google.com/docs/cpp/setup?platform=android#libraries-android-cmake
* iOS -> https://firebase.google.com/docs/cpp/setup?platform=ios#libraries-ios
* desktop: Windows, macOS, and Linux -> https://firebase.google.com/docs/cpp/setup?platform=ios#libraries-desktop

Having this in mind maybe I could already change the `nativeMain` -> `appleMain` & all the `.ios` / `.native` properties that allow us to interact with Firebase SDKs directly to just `.apple`. Then with C++ being added we can see if the structure works or needs some more tweaks. What do you think?

All available targets are defined here: https://github.com/JetBrains/kotlin/blob/92d200e093c693b3c06e53a39e0b0973b84c7ec5/native/utils/src/org/jetbrains/kotlin/konan/target/KonanTarget.kt

suntrix:
@nbransby Any updates?

suntrix:
@nbransby 👋 

I published this to GitHub packages if it helps to get things going (I have more contributions in the pipeline that are blocked by this PR).

Maven repo: `https://maven.pkg.github.com/suntrix/firebase-kotlin-sdk`
Version `1.6.1-kotlin-1-6-10-suntrix-2`

nbransby:
Apologies, we (fingers crossed) will be dedicating more resources to this project in the near future, in the meantime I'll do my best to move things forward.

Looking at FIRPhoneAuthCredential and the APIs round it as an example, I think the breaking change of removing them from the common layer to support the additional apple platforms is probably no the way to go as it will mean these APIs cannot be used from common code for the 90% of use cases (iOS/Android/JS).

What's the reason its not supported on the other apple platforms? If will be in the future I think it would make more sense to throw a NotImplementedException at runtime and keep the API in the common layer. If its permanent and there are more APIs that are missing it might be worth using the hierarchical project structure you linked to with a additional layer to add the missing common APIs that the other apple platform would not inherit from like they have done with runBlocking 

image

suntrix:
Unfortunately I have no idea if any of the iOS-only APIs will ever be supported on other Apple platforms (that's a good question the the Firebase Team), so I think throwing NotImplementedException is not a good idea.

Also their header docs don't say much & are riddled with`API_UNAVAILABLE`...
```
/** @class FIRPhoneAuthCredential
    @brief Implementation of FIRAuthCredential for Phone Auth credentials.
        This class is available on iOS only.
 */
NS_SWIFT_NAME(PhoneAuthCredential) API_UNAVAILABLE(macos, tvos, watchos)
    @interface FIRPhoneAuthCredential : FIRAuthCredential<NSSecureCoding>
```

```
/** @fn sendSignInLinkToEmail:actionCodeSettings:completion:
    @brief Sends a sign in with email link to provided email address.

    @param email The email address of the user.
    @param actionCodeSettings An `FIRActionCodeSettings` object containing settings related to
        handling action codes.
    @param completion Optionally; a block which is invoked when the request finishes. Invoked
        asynchronously on the main thread in the future.
 */
- (void)sendSignInLinkToEmail:(NSString *)email
           actionCodeSettings:(FIRActionCodeSettings *)actionCodeSettings
                   completion:(nullable void (^)(NSError *_Nullable error))completion
    API_UNAVAILABLE(watchos);
```

The hierarchical project structure is always additive, `commonMain` contains the most common APIs, then the next level is `nativeMain`, etc. If we'd like to have the problematic APIs available in the `commonMain` we'll end up with lots of NO-OP implementations to satisfy `expect -> actual`.



----

Also after toying with some more targets... I think all the properties that allow us to interact with Firebase SDKs (`.android`, `.ios`, `.js`) directly should be renamed to something like `.native` / `.nativeSdk` / `.firebaseSdkObject` to be more generic & unified across platform targets. What do you think?

suntrix:
Closing for now as I'll be reworking this a bit.

PR #294, opened 2022-04-22
Add EmailAuthProvider.credentialWithLink()

shepeliev:


firebase-auth/src/commonMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -15,6 +15,7 @@ expect class OAuthCredential : AuthCredential
 
 expect object EmailAuthProvider {
     fun credential(email: String, password: String): AuthCredential
+    fun getCredentialWithLink(email: String, emailLink: String): AuthCredential
```
nbransby, line 18: can we rename to credentialWithLink for consistency

shepeliev: Yes, we can. Thanks for the comment.

nbransby (approved):

PR #297, opened 2022-05-02
[Functions] Fix iOS exception message

walkingbrad:
When an error is returned in a cloud function (e.g. `HttpsError("internal", "This is the error message")`), the `actual` iOS implementation incorrectly propagates the message to `FirebaseFunctionsException`. 

Rather than set `message = message`, it sets `message = theNSError.toString()` which produces an Objective-C string representation of the error. This prevents us from being able to parse that field reliably, since it is in an unstructured format and is inconsistent between Android and iOS.

This change brings iOS in parity with Android, where the exception's `message` matches that of the error returned by Firebase.
e.g. `message = "This is the error message"`


nbransby (approved):

nbransby:
@walkingbrad thanks for the clear description of the issue!

PR #308, opened 2022-05-20
Fix android phone authentication

CharlesEtieve:
Current behavior blocks sms code user input until auto retrieval timeout is reached.

Documentation states it is not recommended to enable user input after sms code retrieval timeout, we should rather enable user input after code sent.
https://firebase.google.com/docs/auth/android/phone-auth#oncodeautoretrievaltimeoutstring-verificationid

So I moved the `getVerificationCode()` from `onCodeAutoRetrievalTimeOut()` to `onCodeSent()` and it works well on my project.


firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -104,6 +101,9 @@ actual class PhoneAuthProvider(val android: com.google.firebase.auth.PhoneAuthPr
                 }
             }
 
+            override fun onCodeAutoRetrievalTimeOut(verificationId: String) {
+            }
```
nbransby, line 105: we'll need to fail the CompletableDeferred here too otherwise it will hang?

firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
```diff
@@ -104,6 +101,9 @@ actual class PhoneAuthProvider(val android: com.google.firebase.auth.PhoneAuthPr
...
+
             override fun onVerificationCompleted(credential: com.google.firebase.auth.PhoneAuthCredential) {
                 response.complete(Result.success(AuthCredential(credential)))
```
nbransby, line 108: i know this isnt your code but we shoudlnt be wrapping the result in a Result - can just fail the CompletableDeferred with [completeExceptionally](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/complete-exceptionally.html)

nbransby (changes requested):
Thanks for the PR! And sorry for the delay in reviewing.

What about JS and iOS, do they suffer the same issue?

firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
Hash3690: I do not think we should make a response.complete() in onCodeAutoRetrievalTimeOut(), because the response can be completed only once and we would not be able to make response.complete() after user input.
Maybe we could add a function inside PhoneVerificationProvider: onCodeAutoRetrievalTimeOut() ?

firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
Hash3690: So instead of response.complete(Result.failure(exception)) we should do response.completeExceptionally(exception) ?

Hash3690:
> Thanks for the PR! And sorry for the delay in reviewing.
> 
> What about JS and iOS, do they suffer the same issue?

Hello :)
Sorry for my late reply !
I cannot tell you about the other platforms but I guess not.
My iOS coworker did not implement authentication with this SDK.
It was a little bit too complex with the suspend function inside PhoneVerificationProvider.
I will tell you later about that.
We do not have a JS app.

firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
nbransby: you don't want to complete, you fail it with the appropriate exception

nbransby:
Thanks for this, and genuine apologies for how long it has been open. It's now ~517 commits behind master and the Android credential handling around it has been rewritten since, so the patch no longer applies. Closing as stale — but the underlying phone-auth issue is being picked up separately, so this hasn't been lost. Sorry again for the delay in coming back to you.

PR #337, opened 2022-09-02
Kotlin 1.7.10

suntrix:


rocketraman:
Trying to build this locally and getting the same error as the CI system. @suntrix any ideas?

PR #349, opened 2023-01-27
Added starting point to move over to cocoapods plugin

Reedyuk:
Kotlin 1.8.20 has a few fixes that allows us to use cocoapods instead of carthage.
Converted all subprojects to use cocoapods plugin instead of carthage.
Added caching(works better when its in main branch)

@nbransby or @michaelprichardson - do you want to review this? its a fairly big change.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,9 +19,6 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '11'
-      - uses: maxim-lobanov/setup-xcode@v1
-        with:
-          xcode-version: '13.4.0'
       - name: Setup versions
         uses: eskatos/gradle-command-action@v1
         with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,6 +19,9 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '11'
+      - uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: '13.4.0'
       - name: Setup versions
         uses: eskatos/gradle-command-action@v1
         with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,8 +28,6 @@ jobs:
           arguments: :updateVersions
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
-      - name: Install Carthage
-        run: brew install carthage
       - name: Publish Firebase App
         uses: eskatos/gradle-command-action@v1
         with:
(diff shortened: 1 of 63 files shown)


nbransby:
LGTM - what is the recommended way to include the firebase-kotlin-sdk and firebase in your xcode project now then?

Reedyuk:
Two ways, either generating the pod file or you could just add like we normally do.
The difference now should be the kmp projects that use this library, you shouldn't have to add the dependency for compilation - there is several issues in GitHub raised around this, framework not found

nbransby (approved):

walkingbrad:
Umm... so all of iOS is now TODO? Is there a plan for that? It was very surprising spend the time updating to the latest release here and find that everything was broken at runtime for iOS.

Reedyuk:
> Umm... so all of iOS is now TODO? Is there a plan for that? It was very surprising to spend the time updating to the latest release here and find that everything was broken at runtime for iOS.

What do you mean everything is todo? It's working. What isn't working for you?

walkingbrad:
@Reedyuk Oh I guess it's only for creating Firebase product classes from a specific FirebaseApp instance (which I use extensively in my app). I refer to all of these: `TODO("Come back to issue")`

Still a very sneaky and surprising change that would've been nice to have more upfront communication about in release notes or even the PR description.

So what was the issue there? Something very tricky? Happy to assist if if it's not :P

Reedyuk:
I've not had chance to look at it, yes sorry was my fault.
Happy for you to put a pr in for the changes. I think it was a slight change to the api in iOS.

Reedyuk:
A lot of time was spent just getting it to work with cocoa pods plug-in and then the pain of compilation 😂

walkingbrad:
Are you suggesting that bridging Gradle, Kotlin, iOS, and Cocoapods is complicated? 😂 😂 😂 

I'll take a dive into it next week maybe.

Reedyuk:
> Are you suggesting that bridging Gradle, Kotlin, iOS, and Cocoapods is complicated? 😂 😂 😂
> 
> I'll take a dive into it next week maybe.

It seems like there is an issue with the ios commoizer.

Screenshot 2023-04-22 at 08 51 08

It might an issue i need to raise with the jetbrains team.

walkingbrad:
Could this fix it or be related? I noticed a warning in my own project that recommended I add this after upgrading to kotlin 1.8.20

`kotlin.mpp.enableCInteropCommonization=true` in gradle settings

https://kotlinlang.org/docs/multiplatform-share-on-platforms.html#share-code-on-all-platforms

walkingbrad:
@Reedyuk I've got a PR here #379 if you want to take a look.

PR #353, opened 2023-02-23
Firebase performance library development

OliverMcB:
Start the development of the firebase performance library with the ability to create, start and stop a trace


firebase-perf/src/iosMain/kotlin/dev/gitlive/firebase/perf/metrics/Trace.kt
```diff
@@ -0,0 +1,14 @@
+package dev.gitlive.firebase.perf.metrics
+
+import cocoapods.FirebasePerformance.FIRTrace
+
+actual class Trace(val ios: FIRTrace?) {
```
Reedyuk, line 5: optional?

Reedyuk: Also probably should be an internal constructor?

firebase-perf/src/androidMain/kotlin/dev/gitlive/firebase/perf/metrics/Trace.kt
```diff
@@ -0,0 +1,10 @@
+package dev.gitlive.firebase.perf.metrics
+
+import com.google.firebase.perf.metrics.Trace
+
+actual class Trace(val android: Trace) {
```
Reedyuk, line 5: same here, internal constructor, then we expose methods via commonMain

Reedyuk (changes requested):

firebase-perf/src/iosMain/kotlin/dev/gitlive/firebase/perf/metrics/Trace.kt
OliverMcB: The FIRTrace returned by traceWithName is nullable. If we wanted to make sure we had a trace we could do something like this in the `Performance` class

```
actual fun newTrace(traceName: String): Trace {

        val trace = FIRPerformance.sharedInstance().traceWithName(traceName)

        checkNotNull(trace)

        return Trace(trace)
    }
```

walkingbrad: It's interesting that the iOS return value is nullable but the Android one is non-null (from what I can tell). [This issue](https://github.com/firebase/firebase-ios-sdk/issues/8583) speaks to some of the nuances of why it might have to be nullable on iOS, though I don't know why those wouldn't apply to Android (maybe Android throws an Exception?) - empty string, string > 100 chars.

Might be worth `assert()`ing here so that it crashes in dev but not prod

Reedyuk: this is a wrapper library so it probably should mimic what the native library does. If it returns optional, then it should deal with it being optional.

PR #359, opened 2023-03-15
[trace-expansion] Add remaining functionality to Trace class

OliverMcB:


firebase-common/src/jsMain/kotlin/dev/gitlive/firebase/externals.kt
```diff
@@ -528,16 +528,24 @@ external object firebase {
...
 
     object performance {
-        interface Performance {
+        open class Performance {
             fun trace(
                 performance: Performance,
                 name: String
-            ): Trace
+            ): PerformanceTrace
         }
```
nbransby, line 536: what reference was used to write this?

OliverMcB: It's defined in this API report file here: https://github.com/firebase/firebase-js-sdk/blob/master/common/api-review/performance.api.md

Can also be found in the index file here: https://github.com/firebase/firebase-js-sdk/blob/master/packages/performance/src/index.ts#L103

nbransby: and in the api docs?

OliverMcB: https://firebase.google.com/docs/perf-mon/custom-code-traces?platform=web#web-version-9

nbransby: you should use this reference: https://firebase.google.com/docs/reference/js/performance.firebaseperformance you can see there is no function called trace on the interface FirebasePerformance

nbransby: anyway thats for web version 9 and I think we are using web version 8?

For 8 there is a function called trace on Performance but the signature is different: https://firebase.google.com/docs/reference/js/v8/firebase.performance.Performance#trace

PR #365, opened 2023-04-07
Failing sealed class list encoding / decoding tests

brahyam:
Hi!,

Thanks a lot for the huge effort of making available this library. 

While using the SDK to store a list of sealed classes I ran into the following error on decoding.
`NullPointerException: null cannot be cast to non-null type kotlin.collections.Map<*, *>`

I managed to track it to `firebase-common/androidMain/_decoders.kt` `FirebaseDecoder.structureDecoder` function. (and also its iOS equivalent)

It happens because the descriptor is expecting a `value` property that was not encoded. I managed to fix it with the following
```kotlin
StructureKind.CLASS, StructureKind.OBJECT, PolymorphicKind.SEALED -> (value as Map<*, *>).let { map ->
        FirebaseClassDecoder(decodeDouble, map.size, { map.containsKey(it) }) { desc, index ->
            val elementName = desc.getElementName(index)
            if (desc.kind is PolymorphicKind && elementName == "value") {
                map
            } else {
                map[desc.getElementName(index)]
            }
        }
    }
```
And similar for the iOS but maybe there is a more elegant way?

I will open an issue with the same comment and link this PR

### DIFF
--- a/firebase-common/src/androidMain/kotlin/dev/gitlive/firebase/_decoders.kt
+++ b/firebase-common/src/androidMain/kotlin/dev/gitlive/firebase/_decoders.kt
@@ -11,7 +11,14 @@ import kotlinx.serialization.descriptors.StructureKind
 
 actual fun FirebaseDecoder.structureDecoder(descriptor: SerialDescriptor): CompositeDecoder = when(descriptor.kind) {
         StructureKind.CLASS, StructureKind.OBJECT, PolymorphicKind.SEALED -> (value as Map<*, *>).let { map ->
-            FirebaseClassDecoder(map.size, { map.containsKey(it) }) { desc, index -> map[desc.getElementName(index)] }
+            FirebaseClassDecoder(map.size, { map.containsKey(it) }) { desc, index ->
+                val elementName = desc.getElementName(index)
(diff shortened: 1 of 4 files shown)


Reedyuk:
Can you fix the build failure please.

Reedyuk (approved):

Reedyuk:
@nbransby

brahyam:
@Reedyuk @nbransby any chance it could be merged soon? 🙏🏼

Reedyuk:
I approved, it's on @nbransby

nbransby (changes requested):
there should be separate tests for encode and decode that use nativeAssertEquals and nativeMapOf like the existing tests - this is because you should be testing that it encodes/decodes to the format expected by firebase native libraries

brahyam:
@Reedyuk @nbransby I split the encode and decode tests now and I used `nativeAssertEquals` and `nativeMapOf` for the encoding as other tests.

nbransby (approved):

nbransby:
great stuff!

Reedyuk:
@brahyam - does this work for you locally? the ci is getting stuck on the ios tests,

PR #312, opened 2022-05-24
Serialization improvements

vpodlesnyak:
This pull request uses serializers API to simplify encoding / decoding and provides more type safe way to operate with Timestamps

PR #371, opened 2023-04-11
improve native serialization and timestamp fixes

vpodlesnyak:
- added GeoPoints
- added native serialization of DocumentReference
- fix for Timestamp extensions
- improve debugging


firebase-common/src/jsMain/kotlin/dev/gitlive/firebase/externals.kt
```diff
@@ -453,6 +453,8 @@ external object firebase {
             fun update(field: FieldPath, value: Any?, vararg moreFieldsAndValues: Any?): Promise<Unit>
             fun delete(): Promise<Unit>
             fun onSnapshot(next: (snapshot: DocumentSnapshot) -> Unit, error: (error: Error) -> Unit): ()->Unit
+
+            fun isEqual(other: DocumentReference): Boolean
```
vpodlesnyak, line 457: proper equality checks help a lot when writing tests in the applications using this SDK

README.md
```diff
@@ -88,6 +88,7 @@ db.collection("cities").document("LA").set(City.serializer(), city, encodeDefaul
 ```
 
 The `encodeDefaults` parameter is optional and defaults to `true`, set this to false to omit writing optional properties if they are equal to theirs default values.
+Using [@EncodeDefault](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/) on properties is a recommended way to locally override the behavior set with `encodeDefaults`.
```
vpodlesnyak, line 91: it helped me a lot so why not to mention it for other devs

firebase-firestore/src/androidMain/kotlin/dev/gitlive/firebase/firestore/GeoPoint.kt
```diff
@@ -0,0 +1,19 @@
+package dev.gitlive.firebase.firestore
+
+import kotlinx.serialization.Serializable
+
+/** A class representing a platform specific Firebase GeoPoint. */
+actual typealias NativeGeoPoint = com.google.firebase.firestore.GeoPoint
+
+/** A class representing a Firebase GeoPoint. */
+@Serializable(with = GeoPointSerializer::class)
```
vpodlesnyak, line 9: this approach is similar to a PR [#312](https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/312)

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/Timestamp.kt
```diff
@@ -32,8 +32,8 @@ expect class Timestamp internal constructor(nativeValue: NativeTimestamp): BaseT
 }
 
 fun Timestamp.Companion.fromMilliseconds(milliseconds: Double): Timestamp =
-    Timestamp((milliseconds / 1000).toLong(), (milliseconds * 1000).toInt() % 1000000)
-fun Timestamp.toMilliseconds(): Double = seconds * 1000 + (nanoseconds / 1000.0)
+    Timestamp((milliseconds / 1000).toLong(), ((milliseconds % 1000) * 1000000).toInt())
```
vpodlesnyak, line 35: a bug fix here. nanoseconds weren't calculated properly. added a test

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -79,27 +79,35 @@ expect open class Query {
...
-fun Query.where(field: String, equalTo: Any?) = _where(field, equalTo)
-fun Query.where(path: FieldPath, equalTo: Any?) = _where(path, equalTo)
-fun Query.where(field: String, equalTo: DocumentReference) = _where(field, equalTo)
-fun Query.where(path: FieldPath, equalTo: DocumentReference) = _where(path, equalTo)
-fun Query.where(field: String, lessThan: Any? = null, greaterThan: Any? = null, arrayContains: Any? = null) = _where(field, lessThan, greaterThan, arrayContains)
-fun Query.where(path: FieldPath, lessThan: Any? = null, greaterThan: Any? = null, arrayContains: Any? = null) = _where(path, lessThan, greaterThan, arrayContains)
-fun Query.where(field: String, inArray: List<Any>? = null, arrayContainsAny: List<Any>? = null) = _where(field, inArray, arrayContainsAny)
-fun Query.where(path: FieldPath, inArray: List<Any>? = null, arrayContainsAny: List<Any>? = null) = _where(path, inArray, arrayContainsAny)
+/** @return a native value of a wrapper or self. */
+private val Any.value get() = when (this) {
```
vpodlesnyak, line 83: this unwrapping is needed to properly support querying by natively supported classes. added a test

vpodlesnyak:
@Reedyuk / @nbransby please take a look when you have time. this PR is a follow up for PR [#312](https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/312). cheers

Reedyuk (approved):

Reedyuk:
@nbransby

firebase-firestore/src/commonTest/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -28,6 +28,11 @@ expect val emulatorHost: String
 expect val context: Any
 expect fun runTest(test: suspend CoroutineScope.() -> Unit)
 
+/** @return a map extracted from the encoded data. */
+expect fun encodedAsMap(encoded: Any?): Map<String, Any?>
+/** @return pairs as raw encoded data. */
+expect fun Map<String, Any?>.asEncoded(): Any
+
```
nbransby, line 35: we have nativeMapOf and nativeAssertEquals  that can be reused instead of adding additional calls

nbransby (changes requested):

firebase-firestore/src/commonTest/kotlin/dev/gitlive/firebase/firestore/firestore.kt
vpodlesnyak: @nbransby thanks, I actually missed those. however in order for `nativeMapOf` and `nativeAssertEquals` to be available in `firebase-firestore` tests, the methods shall be moved from the tests folder of the `firebase-common` to a separate module (lets say `firebase-common-test`). shall I make this change?

nbransby: good point - maybe would be easier to do it on a source set level instead of a separate module?

could add a commonTestUtil source set to firebase-common and include in the test source of firebase-firestore with:

```
        val commonTest by getting {
            kotlin.srcDir("../firebase-common/src/commonTestUtil/kotlin")
        }
```

nbransby:
I've noticed a separate issue regarding the TimestampSerializer, compiling a project with the following class defined:

```
@Serializable
class Data(val timestamp: Timestamp)
```

fails with

```
> Task :compileDebugKotlinAndroid FAILED
e: Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
    class dev.gitlive.firebase.firestore.TimestampSerializer, unresolved supertypes: dev.gitlive.firebase.SpecialValueSerializer
```

Any ideas on that?

vpodlesnyak:
@nbransby the Timestamp gets encoded/decoded in TimestampTests so it looks suspicious. I'll investigate further by creating a sample app

firebase-firestore/src/commonTest/kotlin/dev/gitlive/firebase/firestore/firestore.kt
vpodlesnyak: the suggested approach created a lot of noise - it required setting up not only common but platform tests. extracted a test-utils module instead

vpodlesnyak:
@nbransby the serializer not found issue was caused by the fact `firebase-common` was referenced as `implementation` instead of `api`. added a fix

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/Timestamp.kt
```diff
@@ -84,7 +95,7 @@ object DoubleAsTimestampSerializer : SpecialValueSerializer<Double>(
     toNativeValue = { value ->
         when(value) {
             serverTimestamp -> FieldValue.serverTimestamp.nativeValue
-            else -> Timestamp.fromMilliseconds(value)
+            else -> Timestamp.fromMilliseconds(value).nativeValue
```
vpodlesnyak, line 98: a bug fix - the native value shall be extracted as in the line above

nbransby:
> e: Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:

can you use composition instead of inheritance for the logic shared in SpecialValueSerializer? As making the common module an api import will expose its public api to library consumers but it was designed to hold library internal shared code

vpodlesnyak:
@nbransby 

> can you use composition instead of inheritance for the logic shared in SpecialValueSerializer? As making the common module an api import will expose its public api to library consumers but it was designed to hold library internal shared code

I might be wrong but I think the error is not related to composition/inheritance, but to whether the end user adds a `firebase-common` dependency. So we'd either have to make it `api` or instruct the user to add a `firebase-common` dependency.

Please also take a look at `@FirebaseClassDiscriminator` located in `firebase-common`, it looks like it is meant to be exposed

nbransby:
composition won't expose SpecialValueSerializer in the public API of DoubleAsTimestampSerializer therefore the firebase-common dependency is not required to be api (or manually included by client/end user). The readme doesn't actually mention the firebase-common module and adding FirebaseClassDiscriminator there was not ideal but since its there now its probably best to keep the dependency as api.

Instead, for the firebase-common API not intended to be public we should create something like:
https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-internal-serialization-api/

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
walkingbrad: am I right to assume that this change is required to perform a query that requires providing a timestamp field? 
e.g. `where("endAt", greaterThan = Timestamp.now())`

I am unable to do so in `1.8.1` right now and am hoping this fixes it :)

nbransby (approved):
Thanks for your hard work @vpodlesnyak

nbransby:
I've noticed `gradle build` is failing on master now with:
`e: file:///Users/nbransby/IdeaProjects/firebase-kotlin-sdk/firebase-app/src/commonTest/kotlin/dev/gitlive/firebase/firebase.kt:7:12 Expected property 'context' has no actual declaration in module <firebase-app_debugUnitTest> for JVM`
Do you think that was caused by your changes with test-utils @vpodlesnyak ?
We should really starting `gradle build` in the CI instead of just `gradle assemble`!

vpodlesnyak:
@nbransby 
> Thanks for your hard work @vpodlesnyak

my pleasure

> build is failing

I see that the failure is caused by **unit**Tests and there are no unit test folders in the repo. I think the error was caused by a commit `30cc1f94`. where this bit was removed
```
    tasks.whenTaskAdded {
        enabled = when(name) {
            "compileDebugUnitTestKotlinAndroid" -> false
            "compileReleaseUnitTestKotlinAndroid" -> false
            "testDebugUnitTest" -> false
            "testReleaseUnitTest" -> false
            else -> enabled
        }
    }
```

nbransby:
good spot!

PR #379, opened 2023-05-22
Re-enable ios constructors

walkingbrad:
#349 commented out these constructors due to compilation issues. ~~I think things are compiling fine now, for whatever reason, so I'm just uncommenting them back.~~ Looks like there's just some build-time confusion about `FirebaseApp` types that a force-cast alleviates. Thanks @Daeda88! 

Note: `fun useFunctionsEmulator(origin: String)` has been removed as it is no longer present in the latest Firebase version that this repo is now pointing to (it used to be deprecated).

### DIFF
--- a/firebase-auth/src/iosMain/kotlin/dev/gitlive/firebase/auth/auth.kt
+++ b/firebase-auth/src/iosMain/kotlin/dev/gitlive/firebase/auth/auth.kt
@@ -19,8 +19,7 @@ import platform.Foundation.*
 actual val Firebase.auth
     get() = FirebaseAuth(FIRAuth.auth())
 
-actual fun Firebase.auth(app: FirebaseApp): FirebaseAuth = TODO("Come back to issue")
-//    FirebaseAuth(FIRAuth.authWithApp(app.ios))
+actual fun Firebase.auth(app: FirebaseApp): FirebaseAuth = FirebaseAuth(FIRAuth.authWithApp(app.ios))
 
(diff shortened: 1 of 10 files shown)


walkingbrad:
Error:
```
e: file:///Users/runner/work/firebase-kotlin-sdk/firebase-kotlin-sdk/firebase-auth/src/iosMain/kotlin/dev/gitlive/firebase/auth/auth.kt:22:93 Type mismatch: inferred type is cocoapods.FirebaseCore.FIRApp but objcnames.classes.FIRApp was expected
```

Looks related to https://youtrack.jetbrains.com/issue/KT-56468

Daeda88:
> Error:
> 
> ```
> Type mismatch: inferred type is cocoapods.FirebaseCore.FIRApp but objcnames.classes.FIRApp was expected
> ```
> 
> Looks related to https://youtrack.jetbrains.com/issue/KT-56468

Its essentially a type alias, so you should be able to just cast

walkingbrad:
Looks to be working in my local setup as well! Thanks @Daeda88, I probably never would have tried this. 

I am successfully publishing this lib to mavenLocal and running my iOS app that depends on a few of these iOS constructor APIs to load a dynamic `FirebaseApp` at runtime

I've tested these at runtime, and all the others should work too since the logic I introduced is the same:
- cloud functions
- firestore
- auth

nbransby (approved):
Great stuff, thanks both!

Issue #289, opened 2022-04-10
Handle `FirebaseAuthWeakPasswordException` subclass of `FirebaseAuthInvalidCredentialsException`


rocketraman:
In the Android SDK, `FirebaseAuthWeakPasswordException` extends `FirebaseAuthInvalidCredentialsException` (see https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuthWeakPasswordException).

The corresponding error code is `auth/weak-password` -- see the corresponding code in the JS SDK: https://github.com/firebase/firebase-js-sdk/blob/69ff8eb549e49de51cae11a04bce023bb6e1fc02/packages/auth/src/core/errors.ts#L551.

The current code mapping (https://github.com/GitLiveApp/firebase-kotlin-sdk/blob/master/firebase-auth/src/jsMain/kotlin/dev/gitlive/firebase/auth/auth.kt#L157) will handle this relatively important case as a generic error.

PR #387, opened 2023-06-16
Add support/handling for FirebaseAuthWeakPasswordException

Bradleycorn:
Addresses #289 by adding support for the `FirebaseAuthWeakPasswordException` (https://firebase.google.com/docs/reference/kotlin/com/google/firebase/auth/FirebaseAuthWeakPasswordException) in the **firebase-auth** module.

- Marks `FirebaseAuthInvalidCredentialsException` as `open` so that it can be inherited from (more info below).
- Adds a new class `FirebaseAuthWeakPasswordException` that inherits from `FirebaseAuthInvalidCredentialsException`. This is done for two reasons:
   1.  In Firebase itself, the weak password exception inherits from the invalid credentials exception. This keeps that same structure in this library.
   2. It prevents breaking changes. Existing code in consumer applications that is expecting `FirebaseAuthInvalidCredentialsException` will continue to function normally, since the new `FirebaseAuthWeakPasswordException` IS a `FirebaseAuthInvalidCredentialsException`.
- Implements the `FirebaseAuthWeakPasswordException` for all platforms (android, iOS, js).


Bradleycorn:
@Reedyuk - The `assemble` task on the Pull Request build has been running for 5 hours .. That seems ... not right. Any ideas?

Reedyuk:
I've cancelled it and reran it, there is a pr with kotlin bump to fix it

nbransby (approved):
good stuff, thanks @Bradleycorn

PR #418, opened 2023-09-03
Use es modules

nbransby:


nbransby:
Superseded — the legacy per-module `package.json` files this depended on were removed in #838, so this approach no longer applies.

PR #446, opened 2023-12-16
Wait for emulator to start

MattSkala:
It sometimes happens the tests are executed before Firebase Emulator is ready due to a race condition. Waiting for the Firebase emulators to start should be make tests more reliable.


Daeda88 (approved):

nbransby:
android failed 😞

Daeda88:
I've had current iOS failure, remove the iOS cocoapods cache to fix

MattSkala:
The Android test is flaky, it still sometimes fails, but it shouldn't be made worse by these changes as it happens in other PRs as well. It passed now after restart.

I've cleared cocoapods cache by changing the cache key.

nbransby (approved):

Issue #132, opened 2021-01-04
Update to kotlin 1.4.30 and use inline classes


nbransby:
This would remove the overheads associated with the wrapper classes

PR #448, opened 2023-12-19
Adding Serialization fixes

Daeda88:
The current implementation of serializing classes does not work correctly when dealing with nested classes. 

This PR aims to fix this by rewriting the encoding/decoding. This introduces EncodingSettings and DecodingSettings to Database, Firestore and Functions. On Firestore additional async support has been added


firebase-common/src/commonTest/kotlin/dev/gitlive/firebase/EncodersTest.kt
```diff
@@ -12,125 +12,169 @@ import kotlin.test.assertEquals
 import kotlin.test.assertNull
 import dev.gitlive.firebase.nativeAssertEquals
 import dev.gitlive.firebase.nativeMapOf
+import kotlinx.serialization.builtins.MapSerializer
+import kotlinx.serialization.builtins.serializer
+import kotlinx.serialization.modules.SerializersModule
 
 @Serializable
-data class TestData(val map: Map<String, String>, val bool: Boolean = false, val nullableBool: Boolean? = null)
+object TestObject {
```
Daeda88, line 20: Run these tests on the current project to see why encoding is currently broken

nbransby:
are their any breaking changes in here?

Daeda88:
There are a few where `encodeDefaults: Boolean` was changed to `encodeSettings: EncodeSettings`. On some places I did add convenience extension methods, but that would require an import (and I also forgot in a few places). Ill try and add legacy accessors. This is gonna conflict with the where PR though, so it would be great if we can get that merged first so I can track the changes more easily.

Btw, if you're worried about API changes, consider adding https://github.com/Kotlin/binary-compatibility-validator to the project. We use it for [Kaluga](https://github.com/splendo/kaluga) and it makes it much easier to spot changes

nbransby:
> Btw, if you're worried about API changes, consider adding https://github.com/Kotlin/binary-compatibility-validator to the project. We use it for [Kaluga](https://github.com/splendo/kaluga) and it makes it much easier to spot changes

Yes we need this!

Daeda88:
Made changes to keep API stable

Daeda88:
@nbransby seems publication to maven central failed: https://central.sonatype.com/artifact/dev.gitlive/firebase-java-sdk/versions

nbransby:
> @nbransby seems publication to maven central failed: https://central.sonatype.com/artifact/dev.gitlive/firebase-java-sdk/versions

fixed this now

nbransby:
Found some more time to review...

I think we should keep named arguments instead of the EncodeSettings class, revert https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/449#issuecomment-1873835429 and update the examples to some of the places the new arguments are added. 

Having to wrap settings in EncodeSettings(...) is necessary boilerplate in the public API, people will likely only want to set 1 or 2 settings and named arguments with default values works great for that.

Daeda88:
Done

firebase-common/src/commonMain/kotlin/dev/gitlive/firebase/EncodeDecodeSettings.kt
```diff
@@ -0,0 +1,33 @@
...
+    override val serializersModule: SerializersModule = EmptySerializersModule(),
+) : EncodeDecodeSettings()
+
+/**
+ * [EncodeDecodeSettings] used when decoding an object
+ * @param serializersModule the [SerializersModule] to use for deserialization. This allows for polymorphic serialization on runtime
+ */
+data class DecodeSettings(
+    override val serializersModule: SerializersModule = EmptySerializersModule(),
+) : EncodeDecodeSettings()
```
nbransby, line 33: cant these be internal now?

Daeda88: Some issues with some inline funs.

But actually, I'd like to propose that instead of named arguments we try a builder approach. E.g.:

`set(value) { shouldEncodeDefaults = false }`

Reason im suggesting it is because its annoying as hell right now to add new settings, since you have dozens of methods that need them specified. A builder makes it a one line addition

Daeda88:
Im trying a builder approach instead, as suggested here https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/448#discussion_r1450851712

The advantage of this compared to named arguments is that it is much easier to maintain.

Daeda88:
@nbransby in case you werent aware, the builder approach is in the latest commit. Ive also tracked master again

nbransby:
Should all functions that take a builder lambda  be marked as inline to avoid the lambda allocation?

nbransby:
Whats the point of the Async nested classes and the Deferred.convert function?

Daeda88:
> Should all functions that take a builder lambda be marked as inline to avoid the lambda allocation?
I dont think that actually avoids anything. We can make the lambda nullable I guess

> Whats the point of the Async nested classes and the Deferred.convert function?

Well, to improve async behaviour. For instance, logging or grouping. E.g.
```kotlin
suspend fun deleteSomeDocs() {
    val toExecute = listOf(
       firestore.collectionn("Foo").document("Bar").async.delete()
       firestore.collectionn("Foo").document("BarBar").async.delete()
    )
    println("Started deletion")
    // do some stuff in the mean time
    toExectute.awaitAll()
}
```

Deferred convert is a mapper for deferred. The Android sdk would return an AndroidDocumentReference and you'd want to return a Kotlin DocumentReference, so it should be mapped.

nbransby:
> I dont think that actually avoids anything. We can make the lambda nullable I guess
> 
> >

It does, thats the whole point of inline functions

nbransby:
> Well, to improve async behaviour. For instance, logging or grouping. E.g.

As this is just a generic async utility function and unrelated to firebase we shouldn't be introducing it to our public API, you are welcome to add it to the test sources if it makes writing tests easier

nbransby:
> Deferred convert is a mapper for deferred. The Android sdk would return an AndroidDocumentReference and you'd want to return a Kotlin DocumentReference, so it should be mapped.

Same with this, move it to test sources if you use it in the tests

Daeda88:
@nbransby Ive removed the Async classes and made some changes so all builders funs are now inline. Ive annotated methods with `@PublishedAPI` if only used by the inline methods. The actual implementations of builders are now hidden from the end users.

In addition, Ive made changes to fix `runTransaction` on database, so it now properly uses encode/decode settings

README.md
```diff
@@ -85,13 +85,16 @@ data class City(val name: String)
 Instances of these classes can now be passed [along with their serializer](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#introduction-to-serializers) to the SDK:
 
 ```kotlin
-db.collection("cities").document("LA").set(City.serializer(), city, encodeDefaults = true)
+db.collection("cities").document("LA").set(City.serializer(), city) { shouldEncodeElementDefault = true }
```
nbransby, line 88: think we should keep the name encodeDefaults, as thats how all other formats surface this property to clients
image

README.md
```diff
@@ -85,13 +85,16 @@ data class City(val name: String)
...
 ```
 
-The `encodeDefaults` parameter is optional and defaults to `true`, set this to false to omit writing optional properties if they are equal to theirs default values.
+The `buildSettings` closure is optional and allows for configuring serialization behaviour. 
+
+Setting the `shouldEncodeElementDefault` parameter is optional and defaults to `true`, set this to false to omit writing optional properties if they are equal to theirs default values.
 Using [@EncodeDefault](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/) on properties is a recommended way to locally override the behavior set with `encodeDefaults`.
 
-You can also omit the serializer but this is discouraged due to a [current limitation on Kotlin/JS and Kotlin/Native](https://github.com/Kotlin/kotlinx.serialization/issues/1116#issuecomment-704342452)
+You can also omit the serializer but this is discouraged due to a [current limitation on Kotlin/JS and Kotlin/Native](https://github.com/Kotlin/kotlinx.serialization/issues/1116#issuecomment-704342452).
```
nbransby, line 96: this line can be deleted now as its been fixed

README.md
```diff
@@ -85,13 +85,16 @@ data class City(val name: String)
...
+
+Setting the `shouldEncodeElementDefault` parameter is optional and defaults to `true`, set this to false to omit writing optional properties if they are equal to theirs default values.
 Using [@EncodeDefault](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-encode-default/) on properties is a recommended way to locally override the behavior set with `encodeDefaults`.
 
-You can also omit the serializer but this is discouraged due to a [current limitation on Kotlin/JS and Kotlin/Native](https://github.com/Kotlin/kotlinx.serialization/issues/1116#issuecomment-704342452)
+You can also omit the serializer but this is discouraged due to a [current limitation on Kotlin/JS and Kotlin/Native](https://github.com/Kotlin/kotlinx.serialization/issues/1116#issuecomment-704342452). 
+To support [contextual serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#contextual-serialization) or [open polymorphism](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md#open-polymorphism) the `serializersModule` can be overridden in any `EncodeSettings` or `DecodeSettings`
```
nbransby, line 97: Can you add an example? you talk about `EncodeSettings` or `DecodeSettings` here but above you mention `buildSettings`

firebase-common/src/commonMain/kotlin/dev/gitlive/firebase/reencodeTransformation.kt
```diff
@@ -0,0 +1,22 @@
+package dev.gitlive.firebase
+
+import kotlinx.serialization.KSerializer
+
+inline fun <reified T> reencodeTransformation(value: Any?, builder: EncodeDecodeSettingsBuilder.() -> Unit = {}, transform: (T) -> T): Any? {
```
nbransby, line 5: houls these be public?

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/helpers.kt
```diff
@@ -5,20 +5,20 @@ import kotlin.jvm.JvmName
 
 //** Helper method to perform an update operation. */
 @JvmName("performUpdateFields")
-fun encodeFieldAndValue(
+inline fun encodeFieldAndValue(
```
nbransby, line 8: these should also be internal?

README.md
Daeda88: Renamed

README.md
Daeda88: Removed

README.md
Daeda88: Updated

firebase-common/src/commonMain/kotlin/dev/gitlive/firebase/reencodeTransformation.kt
Daeda88: Yes, just like encode/decode methods, they are to be used by the modules (in this case database) so a public modified is required to access them.

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/helpers.kt
Daeda88: Yeah, moved to internal w @PublishedAPI

Daeda88:
Ive included changes proposed by #456

nbransby:
Looking good, whats the purpose of the Base... and Native... classes such as NativeQuery and s BaseTransaction? @Daeda88

Daeda88:
> Looking good, whats the purpose of the Base... and Native... classes such as NativeQuery and s BaseTransaction? @Daeda88

So with expect/actuals classes you cannot implement code on the common side. Even though with a lot of the code, you kind of want it (e.g., most methods need to encode/decode their values). Base<ClassName> solves this by introducing an abstract class that the expect/actuals classes implement. It makes code much more maintainable and also testable since they can be mocked. Native<Class> is a solution to allow for common constructors. This is useful for the abstract classes (e.g. `BaseDatabaseReference`) that may map the common behaviour. All in all, its to reduce code duplication.

nbransby:
> Even though with a lot of the code, you kind of want it (e.g., most methods need to encode/decode their values). Base solves this by introducing an abstract class that the expect/actuals classes implement.

I assume this is to share code internally within the library? I also assume we don't need to share additional internal state (the only state our wrapper classes should ever hold is the reference to the native instance[^1] in which case inheritance is probably not the right tool here, wouldn't internal extension functions be a better fit? Want to prevent polluting the public API with inheritance chains when their only purpose is internal code sharing.

[^1]: in fact one day it would be nice if our wrapper classes were [compiled away at build time](https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/132))

Daeda88:
> > Even though with a lot of the code, you kind of want it (e.g., most methods need to encode/decode their values). Base solves this by introducing an abstract class that the expect/actuals classes implement.
> 
> I assume this is to share code internally within the library? I also assume we don't need to share additional internal state (the only state our wrapper classes should ever hold is the reference to the native instance[1](#user-content-fn-1-6c88b1acbb299166cc14a4d10e5c5c3e) in which case inheritance is probably not the right tool here, wouldn't internal extension functions be a better fit? Want to prevent polluting the public API with inheritance chains when their only purpose is internal code sharing.
> ## Footnotes
> 
>     1. in fact one day it would be nice if our wrapper classes were [compiled away at build time](https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/132)) [↩](#user-content-fnref-1-6c88b1acbb299166cc14a4d10e5c5c3e)

There's no state being stored whatsoever. This is about common code having common behaviour. To give an example with the changes:

Old:
```kotlin
// Common
expect class DocumentReference internal constructor(nativeValue: NativeDocumentReference) {
    suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean = true, merge: Boolean = false)
    suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean = true, vararg mergeFields: String)
    suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean = true, vararg mergeFieldPaths: FieldPath)

    suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean = true, merge: Boolean = false)
    suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean = true, vararg mergeFields: String)
    suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean = true, vararg mergeFieldPaths: FieldPath)
}

// Android
actual class DocumentReference actual constructor(internal actual val nativeValue: NativeDocumentReference) {
    actual suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, merge: Boolean) = when(merge) {
        true -> android.set(encode(data, encodeDefaults)!!, SetOptions.merge())
        false -> android.set(encode(data, encodeDefaults)!!)
    }.await().run { Unit }

    actual suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, vararg mergeFields: String) =
        android.set(encode(data, encodeDefaults)!!, SetOptions.mergeFields(*mergeFields))
            .await().run { Unit }

    actual suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, vararg mergeFieldPaths: FieldPath) =
        android.set(encode(data, encodeDefaults)!!, SetOptions.mergeFieldPaths(mergeFieldPaths.map { it.android }))
            .await().run { Unit }

    actual suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, merge: Boolean) = when(merge) {
        true -> android.set(encode(strategy, data, encodeDefaults)!!, SetOptions.merge())
        false -> android.set(encode(strategy, data, encodeDefaults)!!)
    }.await().run { Unit }

    actual suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, vararg mergeFields: String) =
        android.set(encode(strategy, data, encodeDefaults)!!, SetOptions.mergeFields(*mergeFields))
            .await().run { Unit }

    actual suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, vararg mergeFieldPaths: FieldPath) =
        android.set(encode(strategy, data, encodeDefaults)!!, SetOptions.mergeFieldPaths(mergeFieldPaths.map { it.android }))
            .await().run { Unit }
}

// iOS
actual class DocumentReference actual constructor(internal actual val nativeValue: NativeDocumentReference) {
    actual suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, merge: Boolean) =
        await { ios.setData(encode(data, encodeDefaults)!! as Map<Any?, *>, merge, it) }

    actual suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, vararg mergeFields: String) =
        await { ios.setData(encode(data, encodeDefaults)!! as Map<Any?, *>, mergeFields.asList(), it) }

    actual suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, vararg mergeFieldPaths: FieldPath) =
        await { ios.setData(encode(data, encodeDefaults)!! as Map<Any?, *>, mergeFieldPaths.map { it.ios }, it) }

    actual suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, merge: Boolean) =
        await { ios.setData(encode(strategy, data, encodeDefaults)!! as Map<Any?, *>, merge, it) }

    actual suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, vararg mergeFields: String) =
        await { ios.setData(encode(strategy, data, encodeDefaults)!! as Map<Any?, *>, mergeFields.asList(), it) }

    actual suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, vararg mergeFieldPaths: FieldPath) =
        await { ios.setData(encode(strategy, data, encodeDefaults)!! as Map<Any?, *>, mergeFieldPaths.map { it.ios }, it) }
}
```

All these expect/actuals funs can and should be split up into two parts: 
1) A common behaviour, the encoding.
2) A platform specific behaviour, here set data

Using abstract classes to do part 1 in common code, and then forwarding the result to 2 to do something expect/actuals reduces the footprint greatly:

```kotlin
// Common
abstract class BaseDocumentReference {
    suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, merge: Boolean = false) = setEncoded(encode(data, encodeDefaults)!!, if (merge) SetOptions.Merge else SetOptions.Overwrite)
    suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, vararg mergeFields: String) = setEncoded(encode(data, encodeDefaults)!!, SetOptions.MergeFields(mergeFields.asList()))
    suspend inline fun <reified T> set(data: T, encodeDefaults: Boolean, vararg mergeFieldPaths: FieldPath) = setEncoded(encode(data, encodeDefaults)!!, SetOptions.MergeFieldPaths(mergeFieldPaths.asList()))

suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, merge: Boolean = false) = setEncoded(
        encode(strategy, data, encodeDefaults)!!, if (merge) SetOptions.Merge else SetOptions.Overwrite)
suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, vararg mergeFields: String) = setEncoded(
        encode(strategy, data, encodeDefaults)!!, SetOptions.MergeFields(mergeFields.asList()))
suspend fun <T> set(strategy: SerializationStrategy<T>, data: T, encodeDefaults: Boolean, vararg mergeFieldPaths: FieldPath) = setEncoded(
        encode(strategy, data, encodeDefaults)!!, SetOptions.MergeFieldPaths(mergeFieldPaths.asList()))

    @PublishedApi
    internal abstract suspend fun setEncoded(encodedData: Any, setOptions: SetOptions)
}

expect class DocumentReference internal constructor(nativeValue: NativeDocumentReference) : BaseDocumentReference

// Android
actual class DocumentReference actual constructor(internal actual val nativeValue: NativeDocumentReference) : BaseDocumentReference() {
    override suspend fun setEncoded(encodedData: Any, setOptions: SetOptions) {
        val task = (setOptions.android?.let {
            android.set(encodedData, it)
        } ?: android.set(encodedData))
        task.await()
    }
}

// iOS
actual class DocumentReference actual constructor(internal actual val nativeValue: NativeDocumentReference) : BaseDocumentReference() {
    override suspend fun setEncoded(encodedData: Any, setOptions: SetOptions) = await {
        when (setOptions) {
            is SetOptions.Merge -> ios.setData(encodedData as Map<Any?, *>, true, it)
            is SetOptions.Overwrite -> ios.setData(encodedData as Map<Any?, *>, false, it)
            is SetOptions.MergeFields -> ios.setData(encodedData as Map<Any?, *>, setOptions.fields, it)
            is SetOptions.MergeFieldPaths -> ios.setData(encodedData as Map<Any?, *>, setOptions.encodedFieldPaths, it)
        }
    }
}
```

Consider how much more maintainable this is. We encode consistently on a single place, and we send to the platform logic in a single place. Before, for these 3 methods, we encoded for each method per platform (so 3 times) and for each method we had to write the same coroutine logic with error handling (also 3 for each method).

You can't do this with extension functions, because that would defeat the purpose where the common step still needs to be implemented on each method per platform.

nbransby:
> You can't do this with extension functions, because that would defeat the purpose where the common step still needs to be implemented on each method per platform.

Are you saying this pattern is not possible?

*common*
```kotlin
expect class Something {
  expect publicApi()
}

internal fun Something.sharedFunction() {}
```

*platform*
````
actual class Something() {
  actual publicApi() {
    sharedFunction()
    //do other platform specific stuff
  }
}

**Comment by Daeda88** [comment]
No, im saying its not correct to do it like that in my opinion, for maintainability reasons. Because if you use that approach, every time something changes to sharedFunction you have to account for it in 3 times as many places.

**Comment by Daeda88** [comment]
At the very least, we should have
common
```
expect class Something {
  expect fun publicApi(parameter: String): Result
}

internal inline fun Something.sharedFunction(parameter: String, parsedMethod: (Any) -> Result): Result
```

platform
```
actual class Something() {
  actual fun publicApi(parameter: String) = sharedFunction(parameter) { any ->
     any.toResult()
  } 
}
```

**Comment by Daeda88** [comment]
If you want to account for value classes, I would propose to move to interfaces with default implementations, as that sort of meets both criteria (maintainability and maintaining possibility to inline)
https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS45LjIyIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiLyoqXG4gKiBZb3UgY2FuIGVkaXQsIHJ1biwgYW5kIHNoYXJlIHRoaXMgY29kZS5cbiAqIHBsYXkua290bGlubGFuZy5vcmdcbiAqL1xuXG5pbXBvcnQga290bGlueC5jb3JvdXRpbmVzLipcblxuaW50ZXJmYWNlIFRlc3Qge1xuICAgIHZhbCBzdHJpbmc6IFN0cmluZ1xuICAgIGZ1biB0ZXN0RnVuKCk6IFN0cmluZyB7IHJldHVybiBcInRlc3Qkc3RyaW5nXCIgfVxufVxuXG5ASnZtSW5saW5lXG52YWx1ZSBjbGFzcyBUZXN0Q2xhc3Mob3ZlcnJpZGUgdmFsIHN0cmluZzogU3RyaW5nKSA6IFRlc3RcblxuZnVuIG1haW4oKSB7XG4gICAgcHJpbnRsbihUZXN0Q2xhc3MoXCJIZWxsb1dvcmxkXCIpLnRlc3RGdW4oKSlcbn0ifQ==

**Comment by nbransby** [comment]
> No, im saying its not correct to do it like that in my opinion, for maintainability reasons. Because if you use that approach, every time something changes to sharedFunction you have to account for it in 3 times as many places.

I'm a little confused here as I don't see how inheritance would make it any better, for reference, here's the inheritance version of my trivial example, as you can see the actual implementation doesn't change but we end up publishing BaseToShareStuffBetweenTheActualImplementations in our public API.

*common*
```kotlin
expect class Something : BaseToShareStuffBetweenTheActualImplementations {
  expect fun publicApi()
}

abstract class BaseToShareStuffBetweenTheActualImplementations {
  internal fun sharedFunction() {}
}

```

*platform*
````
actual class Something() {
  actual fun publicApi() {
    sharedFunction()
    //do other platform specific stuff
  }
}

**Comment by nbransby** [comment]
> If you want to account for value classes, I would propose to move to interfaces with default implementations, as that sort of meets both criteria (maintainability and maintaining possibility to inline)

Definitely an option but we should hold fire on a refactor like this until the expect/acual language feature is stabilized

**Comment by Daeda88** [comment]
You example is the reverse of what we need though.
It would be:

common
```
expect class Something : BaseToShareStuffBetweenTheActualImplementations()

abstract class BaseToShareStuffBetweenTheActualImplementations {
  fun publicApi() {
          doSomethingCommon()
          platformFunction()
    }
  abstract fun platformFunction()
}
```

platform
```
actual class Something() : BaseToShareStuffBetweenTheActualImplementations() {
  override fun platformFunction() {
    //do other platform specific stuff
  }
}
```
There's a big difference. The abstract class serves _only_ to expose the public API methods and handle default implementation for it. It then splits up the public API method in something that can be done in common and something that must be done on the platform.

**Comment by Daeda88** [comment]
To compare:
```
// Common
expect class TestClass {
    fun publicAPI()
}

internal fun TestClass.sharedFunction() {}

// Platform
actual class TestClass {
    actual fun publicAPI() {
        sharedFunction()
    }
}
```

Gives
```
public final class TestClass {
	public fun <init> ()V
	public final fun publicAPI ()V
}
```

Whereas
```
// Common
abstract class BaseTestClass {
    fun publicApi() {
        platformFunction()
    }
    internal abstract fun platformFunction()
}

expect class TestClass : BaseTestClass

// Platform
actual class TestClass : BaseTestClass() {
    override fun platformFunction() {}
}
```

Gives us
```
public abstract class BaseTestClass {
	public fun <init> ()V
	public final fun publicApi ()V
}

public final class TestClass : BaseTestClass {
	public fun <init> ()V
}
```

While its true this exposes the `BaseTestClass` as well, you shoud keep in mind that we dont need to do
```
actual fun publicApi() { 
   sharedFunction()
    //do other platform specific stuff
}
```

in the code for _every single platform_. That is the benefit your get from this. I've had to make changes to nearly all functions related to encoding for these changes. Thats literally dozens if not hundreds of methods. An abstract class pattern reduces the number of places where you need to apply changes by 66%.

**Comment by nbransby** [comment]
> There's a big difference. The abstract class serves _only_ to expose the public API methods and handle default implementation for it. It then splits up the public API method in something that can be done in common and something that must be done on the platform.

Ok I understand now! So in that case you could just do:

*common*
```kotlin

class Something {
  fun publicApi() {
        doSomethingCommon()
        platformFunction()
    }
}

expect internal fun Something.platformFunction()

```

*platform*
```kotlin
actual fun Something.platformFunction() {
    //do other platform specific stuff
}
```

No inheritance needed.

nbransby:
> But I guess this causes an issue due to not having the native instance property (ios, android, js) in the class

Which really can only be solved with what you have implemented. One proposal though, since the base classes contain the public API why don't we call those the original names (eg DatabaseReference)and then make the platform specific subclasses internal (eg JsDatabaseReference)?

There is still the issue of the publicly accessible android, ios, and js properties but they could be implemented as follows:

database.js.kt
```kotlin
val DatabaseReference.js get() = (this as JsDatabaseReference).js
```

Daeda88:
> > But I guess this causes an issue due to not having the native instance property (ios, android, js) in the class
> 
> Which really can only be solved with what you have implemented. One proposal though, since the base classes contain the public API why don't we call those the original names (eg DatabaseReference)and then make the platform specific subclasses internal (eg JsDatabaseReference)?
> 
> There is still the issue of the publicly accessible android, ios, and js properties but they could be implemented as follows:
> 
> database.js.kt
> 
> ```kotlin
> val DatabaseReference.js get() = (this as JsDatabaseReference).js
> ```

Glad you get my considerations. Im happy to do the renames, but keep in mind that this makes the upgrade less frictionless since people would have to import the `js` extension method. If thats fine by you, ill make the renames somewhere in the coming week

nbransby:
> but keep in mind that this makes the upgrade less frictionless since people would have to import the `js` extension method

true, means its a breaking change and we'll need to bump the major version number and change the `js`, `android` and `ios properties on all classes to extensions.  The classes that haven't had a base class introduced could do this:

```kotlin
actual class Database(internal val android: com.firebase.Database)

val Database.android get() = android
```

Daeda88:
@nbransby finally got around to looking at this. I removed abstract classes in favour of expected wrappers.

so instead of
```kotlin
abstract class BaseSomething {
   fun doSomething() {
       doCommon()
       doPlatformSpecific()
   }
   internal abstract fun doPlatformSpecific()
}

expect class Something : BaseSomething
```

I now have

```kotlin
internal expect class NativeSomething {
   fun doPlatformSpecific()
}

class Something internal constructor(internal val native: NativeSomething) {
  fun doSomething() {
       doCommon()
       native.doPlatformSpecific()
   }
}
```

nbransby (approved):

nbransby:
@Daeda88 I believe this might of introduced a bug in the rtdb on js - passing a map to `updateChildren` with `null` values would delete those keys from the rtdb if they exist, this still appears to work on android/jvm but for js it causes permission denied with a rule such as this:
```
          "lastActivity": { ".validate": "!newData.exists() || newData.isNumber()" }
```

My guess is that the serialization changes have caused `null` instead of `undefined` to be passed to the firebase-js-sdk causing it to attempt to set the value to `null` instead of removing the key.

PR #482, opened 2024-03-28
Add functions for creating Firestore for a given database

AKJAW:
The current top level functions don't allow creating Firestore for specific databases in the same Google Cloud project
Screenshot 2024-03-28 at 07 47 06

It is possible by having some work-around using expect actual but it's a big hassle and frakly I wasn't able to make pods cooperate with me
Screenshot 2024-03-28 at 07 45 26
Screenshot 2024-03-28 at 07 45 40

This PR "tries" to introduce those functions however I'm not sure about the JS implementations since I wasn't able to get the Repository to synchronize.


nbransby:
Thanks @AKJAW, could you add a unit test for it? Then we can confirm if it works in JS

AKJAW:
> Thanks @AKJAW, could you add a unit test for it? Then we can confirm if it works in JS

The problem is that I cannot get the repository to synchronize properly, but if you could point me to some existing tests I can try to create something

nbransby:
Existing tests are here: https://github.com/GitLiveApp/firebase-kotlin-sdk/blob/master/firebase-firestore/src/commonTest/kotlin/dev/gitlive/firebase/firestore/firestore.kt

AKJAW:
> Existing tests are here: https://github.com/GitLiveApp/firebase-kotlin-sdk/blob/master/firebase-firestore/src/commonTest/kotlin/dev/gitlive/firebase/firestore/firestore.kt

Are there any existing tests for initialization? All I see that it is used as the set-up for others tests. Do you just want me to create a test which uses these functions and verifies that there is no exception?

nbransby:
> Do you just want me to create a test which uses these functions and verifies that there is no exception?

sounds good

AKJAW:
> > Do you just want me to create a test which uses these functions and verifies that there is no exception?
> 
> sounds good

i've added the tests, however I'm not sure if the one that doesn't specify app will be able to pass, since it probably won't be able to retrieve the app from a global context

nbransby:
Yes the JS tests are failing with
```
e: file:///home/runner/work/firebase-kotlin-sdk/firebase-kotlin-sdk/firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt:64:33 Overload resolution ambiguity: 

> Task :firebase-firestore:compileKotlinJs FAILED
public external fun getFirestore(app: FirebaseApp? = ...): Firestore defined in dev.gitlive.firebase.firestore.externals in file firestore.kt
public external fun getFirestore(app: FirebaseApp? = ..., database: String = ...): Firestore defined in dev.gitlive.firebase.firestore.externals in file firestore.kt
public external fun getFirestore(database: String = ...): Firestore defined in dev.gitlive.firebase.firestore.externals in file firestore.kt
```

nbransby:
The android tests are also failing but the test report for firestore appear to be missing, im not sure why, best to try running them locally

AKJAW:
> The android tests are also failing but the test report for firestore appear to be missing, im not sure why, best to try running them locally

Hello, we decided to not use this library, and I still wasn't able to get the project running locally, so I guess I'll close this pull request / convert it to a feature request for the future

PR #491, opened 2024-04-10
Adds an optional Source argument to DocumentReference and Query

### LINKED ISSUE #263: Add DocumentReference.get(source: Source = Source.DEFAULT) to firebase-firestore for iOS/Android/JS
In all Firestore clients there is an option of passing source when trying to obtain current snapshot of document. This makes it possible to force reading from server.

| Library               | Class                            | Member                                        | Platforms                  |
| :-----------------| :------------------------ | :----------------------------------- | :---------------------- |
| firestore | DocumentReference | get(server: Server)  | Android, iOS, JS  |

This could be introduced in the same way it's available in Android client:

- `Source` enum
- optional argument `source: Source = Source.DEFAULT` of `DocumentReference.get()`

---

#### Underlying API

- Objective C client has [method that takes `Source` enum]((https://firebase.google.com/docs/reference/ios/firebasefirestore/api/reference/Classes/FIRDocumentReference#/c:objc(cs)FIRDocumentReference(im)getDocumentWithSource:completion:))
- JS client has three separate methods:
    - [getDocFromCache(reference)](https://firebase.google.com/docs/reference/js/firestore_.md#getdocfromcache) - behaves as `get(Source.CACHE)`
    - [getDocFromServer(reference)](https://firebase.google.com/docs/reference/js/firestore_.md#getdocfromserver) - behaves as `get(Source.SERVER)
    - [getDoc(reference)](https://firebase.google.com/docs/reference/js/firestore_.md#getdoc) - behaves as `get(Source.DEFAULT)`
- Android client has [method that takes `Source` enum](https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/DocumentReference#get(com.google.firebase.firestore.Source))

---

I can help with implementation but I don't have experience with iOS pipeline (nor I have iOS device to test the implementation)

Qw4z1:
First draft of adding a Source enum to be able to run queries on cached documents instead of waiting for server responses (avoids extra reads) or force fetching from server (ignoring cache).

Note: I haven't been able to test the js implementation as my environment seems to be broken.

[Documentation](https://firebase.google.com/docs/firestore/query-data/get-data#source_options)

Closes #263

### DIFF
--- a/firebase-firestore/src/androidMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
+++ b/firebase-firestore/src/androidMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
@@ -8,6 +8,7 @@ package dev.gitlive.firebase.firestore
 import com.google.firebase.firestore.MetadataChanges
 import dev.gitlive.firebase.Firebase
 import dev.gitlive.firebase.FirebaseApp
+import dev.gitlive.firebase.firestore.Source.*
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
@@ -171,8 +172,8 @@ internal actual class NativeDocumentReference actual constructor(actual val nati
 
     actual fun collection(collectionPath: String) = NativeCollectionReference(android.collection(collectionPath))
 
-    actual suspend fun get() =
-        NativeDocumentSnapshot(android.get().await())
+    actual suspend fun get(source: Source) =
+        NativeDocumentSnapshot(android.get(source.toAndroidSource()).await())
 
     actual suspend fun setEncoded(encodedData: Any, setOptions: SetOptions) {
         val task = (setOptions.android?.let {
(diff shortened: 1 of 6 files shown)


nbransby:
Look good, just need to add a test and have the CI turn green

Daeda88 (approved):

Qw4z1:
Added tests in a separate file to be able to toggle persistence per test. Curious to know if there is a better way to do this though...

Daeda88 (approved):

nbransby:
The android tests for auth failed with `Failed to connect to /10.0.2.2:9099`, not sure why. Also there don't appear to be any js or ios test report artifacts for the build so I'm not even sure those tests ran. @Daeda88 you probably have more insight on these than me, are the ios and js tests running correctly?

Daeda88:
Auth test is unstable on Android for ci due to emulator speed I think. I'd just run it manually to be sure

nbransby:
what about ios and js, ideally I would like to know the new test ran and passed

Qw4z1:
@nbransby looks like the PR workflow has `        if: failure()` as a condition for both upload steps. I'm no expert at this, but I assume that means that artifacts and reports are only uploaded if a test has failed, right? 🤔

Daeda88:
They run on different machines, since those jobs are green, Im pretty sure the tests ran to completion

PR #504, opened 2024-05-04
[BREAKING CHANGE] refactor .js, .android, .ios properties extensions

nbransby:


Reedyuk:
This should pass when its completed the run. Can you check it out @nbransby - incase i missed anything.

Reedyuk:
ok its ready now

PR #511, opened 2024-05-19
Wasm

Thaerith:


arieftb:
is this changes work for wasm module?

Issue #517, opened 2024-05-29
Add Firebase Analytics for [KMP]


DenisPopkov:

atonamy:
Yes, we absolutely need it

DenisPopkov:
I wrote my own solution via Firebase REST API. You can check it out in article — https://habr.com/ru/articles/818891/

atonamy:
> @atonamy I wrote my own solution via Firebase REST API. You can check it out in article — https://habr.com/ru/articles/818891/

According to this article implementation looks pretty simple and straightforward I'm surprised why is still not part of firebase-kotlin-sdk ? 

I wonder if this secret key can be retrieved somehow from this firebase-kotlin-sdk itself?

DenisPopkov:
@atonamy 
> I wonder if this secret key can be retrieved somehow by using tools of this firebase-kotlin-sdk itself?
You need to obtain a secret key for each application within your Firebase console project and hardcode it either inside BuildConfig or in GitHub secret keys, for instance. The secret key will not change regardless. By using firebase-kotlin-sdk you can't get secret keys.

atonamy:
@DenisPopkov 
> @atonamy
> 
> > I wonder if this secret key can be retrieved somehow by using tools of this firebase-kotlin-sdk itself?
> 
> You need to obtain a secret key for each application within your Firebase console project and hardcode it either inside BuildConfig or in GitHub secret keys, for instance. The secret key will not change regardless. By using firebase-kotlin-sdk you can't get secret keys.

Then how firebase-kotlin-sdk internally call other Firebase API without exposing secret key?

DenisPopkov:
@atonamy, google-services.json file contains all the necessary information for making requests to the SDK. However, for the REST API, you would need secret keys, which can be found in the official documentation — https://developers.google.com/analytics/devguides/collection/protocol/ga4/sending-events?hl=ru&client_type=firebase

atonamy:
> [enisPopkov](/DenisPopkov)
@DenisPopkov 
how to use google-services.json instead with KMP any example for Analytics? Using secret key is not a solution for us.

DenisPopkov:
@atonamy 
You can still use google-services.json for all Firebase services for iOS and Android. However, for Desktop, you can't do this as Firebase doesn't support it natively. If you're only using Analytics in your project, you don't need google-services.json at all.

> Using secret key is not a solution for us

[...]

Issue #114, opened 2020-12-11
Sending events


PokerAsh:
Hello, I would like to know if there is already implemented sending events into firebase console? Or is in plans?

I am seeing that you've already working on analytics, but couldn't find any way of calling them. Would be great if it was in these libraries!

Thanks in advance

nbransby:
Hi @PokerAsh no there hasn't been any work done on analytics yet but we do want to add them at some point in the future and PRs are always welcome!

mustafaozhan:
@nbransby Really interested on analytics implementation 👍 but couldn’t find any issue to follow the progress do you have one or should we subscribe to this issue ?

JenniferCabotage:
Very interested in analytics support!

mattdornfeld:
Same!

NathanFallet:
Interested as well!

florianNgbala:
Interested as well !

PR #525, opened 2024-06-06
Add firebase analytics

BasBuijsen:
This PR contains most of the firebase analytics implementation for Android, iOS and JS. This was requested in [this recent issue](https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/517) and [this very old issues](https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/114).

For testing, this is very difficult. There is no real way to ensure logs arrived at the firebase side. There is no emulator for analytics and there is no response from the logEvent call to know if it succeeded. For now i only added a test that ensures the analytics object is available and makes sure it doesn't crash when calling the logEvent function. If you have ideas on how to better test this, please let me know!

To create the module i copy pasted one of the previously existing modules. I think i went through all the steps but im not 100% sure i did everything since there is no real guide for adding a new module. Could you let me know wether i added it in all the right places?


DenisPopkov (approved):

DenisPopkov (approved):

BasBuijsen:
@nbransby this is ready for review now. Could you take a look at it when you have the time?

DenisPopkov (approved):

BasBuijsen:
Functionally everything in this pr works, however the ios tests still fail due to some issue with importing the analytics pod dependency in test runs. I cant seem to figure out what the issue is and i have opened an issue on the firebase ios sdk repository. Until i find a working fix this pr cannot be merged.

BasBuijsen:
@nbransby can this be merged? Since the android tests are still flaky :)

nbransby (approved):

PR #528, opened 2024-06-10
PhoneAuthProvider improvement for ios

kiiskristo:
FIRAuthUIDelegateProtocol is not available outside of the firebase-kotlin-sdk library, so there is no way to inject PhoneVerificationProvider to verifyPhoneNumber method. That makes PhoneAuthProvider impossible to use on IOS platform.


kiiskristo:
I could also update this PR to just have FIRAuthUIDelegateProtocol as optional and have less changes in code:
`val delegate: FIRAuthUIDelegateProtocol?`

Instead of: 
`val delegate: Any?`

That UIDelegate is mostly for viewcontroller based architecture and most of people use SwiftUI nowadays.

BasBuijsen:
Im not a maintainer or anything on this project but isnt is better to add an interface that has the same functions the delegate has? Then you can use that type for the delegate variable and can create a wrapper class for the custom delegate class here that will automatically call the underlying firebase delegate itself.

Just sharing my idea here, this may turn out to be harder than i explain so take it with a grain of salt :)

Daeda88:
`FIRAuthUIDelegateProtocol` should just be a typealias for `FIRAuthUIDelegate` on the iOS side. You are right however that it can be null, so it should probably be a typealias for that

kiiskristo:
@BasBuijsen - thank you, actually ended up doing similar way as you described for now.
@Daeda88 - updated PR to alternative way, looks cleaner and reflects native IOS SDK with optional delegate.

Daeda88 (approved):

Varun-Sethi-Dev:
Can Some one guide me on how to use this Phone Authentication i am able to send the otp but how to do verification

PR #538, opened 2024-06-20
Added jvm test

Reedyuk:


Reedyuk:
@nbransby - would you expect the jvm tests to pass?

PR #539, opened 2024-06-20
Test documentation

Reedyuk:


BasBuijsen:
The path to the html is incorrect. It should be `./build/dokka/htmlMultiModule`.

Issue #534, opened 2024-06-16
Add running the binary compatibility validator to GitHub Actions


nbransby:
https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#use-the-binary-compatibility-validator

PR #550, opened 2024-06-25
Api Validation, Explicit API Mode & Linting

### LINKED ISSUE #535: Use explicit API mode
With explicit API mode, you must:

Add visibility modifiers to your declarations to make them public, instead of relying on the default public visibility. This ensures that you've considered what you're exposing as part of the public API.

Define the types for all your public functions and properties to prevent unintended changes to your API from inferred types.

https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors

Daeda88:
Fixes #535 #534 and adds lint. Also did some cleanup based on code inspection

By its very nature this is a large PR with many code changes. However, no functional changes where made in the code. I recommend reviewing the gradle & workflow scripts and just scanning the actual code

Also, because I know its controversial: I left lint to use the standard Intellij rules. This means that [trailing commas](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas) have been added. I've not disabled any rules except for wildcard imports / import ordering since those tend to conflict with the IDE in my experience


nbransby:
Next step will be to remove the internal classes from the API, I think its just the Native... classes found in these two modules:

https://github.com/GitLiveApp/firebase-kotlin-sdk/blob/b72a3cf86debf4e26b95cb9b2e9ac664157a0036/firebase-firestore/api/android/firebase-firestore.api
https://github.com/GitLiveApp/firebase-kotlin-sdk/blob/b72a3cf86debf4e26b95cb9b2e9ac664157a0036/firebase-database/api/android/firebase-database.api

Daeda88:
As in, you want me to do that in this pr? For reviewability I'd recommend doing it in a separate pr

nbransby:
Agreed

nbransby (approved):

PR #553, opened 2024-06-26
Changes the use of Intervals/Timestamps as long into Kotlin Duration/Instant

Daeda88:
Fixes a bit of a pet peeve to me: When dealing with Intervals and Timestamps, on Kotlin we should be using Kotlin Duration and Kotlin DateTime Instant instead of using Longs.

My reasoning behind this:
- Duration gives annotation to the meaning of a long. No longer do you need to check documentation what the long means (or check if we're using seconds or milliseconds)
- The NSTimeInterval is in seconds while Android/JS communicate in milliseconds. This already lead to several bugs that I've solved in this PR. Using duration makes making mistakes less likely
- If you want to set some timeout to say 3 hours, you no longer need to do calculations like `3 * 60 * 60 * 1000` as you can just pass `3.hours`. This leads to better usage of the library

### DIFF
--- a/firebase-analytics/api/android/firebase-analytics.api
+++ b/firebase-analytics/api/android/firebase-analytics.api
@@ -7,6 +7,7 @@ public final class dev/gitlive/firebase/analytics/AnalyticEventConstantsKt {
 public final class dev/gitlive/firebase/analytics/AnalyticsKt {
 	public static final fun logEvent (Ldev/gitlive/firebase/analytics/FirebaseAnalytics;Ljava/lang/String;Lkotlin/jvm/functions/Function1;)V
 	public static final fun setConsent (Ldev/gitlive/firebase/analytics/FirebaseAnalytics;Lkotlin/jvm/functions/Function1;)V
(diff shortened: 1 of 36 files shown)


Daeda88:
> Durations aren't designed to be used as timestamps, we should probably be using https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/-instant/

Right, though I guess the ticket is also a bit of a misnomer as most of these are intervals not timestamps, for which duration is the right class.

`Firebase.storage.setMaxOperationRetryTimeMillis(10000)` -> `Firebae.storage.setMaxOperationRetryTime(10.seconds)`

to me is the main reason for changing this (well that and is confusion. In the code on master, that 10000 was not divided by 1000 so it would be set to 10000 seconds on iOS)

nbransby:
Agree the timeouts should be Durations

Daeda88:
Ok, there was one place where Instant made more sense (in the config module). Moved to it there

nbransby (approved):

PR #555, opened 2024-06-26
Remove Native classes from Public API

Daeda88:
Fixes based on comment https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/550#issuecomment-2190589666

### DIFF
--- a/firebase-database/api/android/firebase-database.api
+++ b/firebase-database/api/android/firebase-database.api
@@ -38,7 +38,6 @@ public final class dev/gitlive/firebase/database/DataSnapshot {
 public final class dev/gitlive/firebase/database/DatabaseReference : dev/gitlive/firebase/database/Query {
 	public final fun child (Ljava/lang/String;)Ldev/gitlive/firebase/database/DatabaseReference;
 	public final fun getKey ()Ljava/lang/String;
-	public final fun getNativeReference ()Ldev/gitlive/firebase/database/NativeDatabaseReference;
 	public final fun onDisconnect ()Ldev/gitlive/firebase/database/OnDisconnect;
 	public final fun push ()Ldev/gitlive/firebase/database/DatabaseReference;
 	public final fun removeValue (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
@@ -47,9 +46,11 @@ public final class dev/gitlive/firebase/database/DatabaseReference : dev/gitlive
 	public final fun setValue (Lkotlinx/serialization/SerializationStrategy;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public final fun setValue (Lkotlinx/serialization/SerializationStrategy;Ljava/lang/Object;ZLkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun setValue$default (Ldev/gitlive/firebase/database/DatabaseReference;Lkotlinx/serialization/SerializationStrategy;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
+	public final fun setValueEncoded (Ljava/lang/Object;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public final fun updateChildren (Ljava/util/Map;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public final fun updateChildren (Ljava/util/Map;ZLkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun updateChildren$default (Ldev/gitlive/firebase/database/DatabaseReference;Ljava/util/Map;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
+	public final fun updateEncodedChildren (Ldev/gitlive/firebase/internal/EncodedObject;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
 public final class dev/gitlive/firebase/database/FirebaseDatabase {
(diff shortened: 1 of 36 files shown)


firebase-database/api/android/firebase-database.api
```diff
@@ -68,41 +69,17 @@ public final class dev/gitlive/firebase/database/FirebaseDatabase {
...
-	public final fun removeValue (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public final fun setValue (Ljava/lang/Object;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public final fun updateEncodedChildren (Ldev/gitlive/firebase/internal/EncodedObject;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-}
-
 public final class dev/gitlive/firebase/database/OnDisconnect {
 	public final fun cancel (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public final fun getNative ()Ldev/gitlive/firebase/database/NativeOnDisconnect;
 	public final fun removeValue (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public final fun setEncodedValue (Ljava/lang/Object;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
nbransby, line 75: why have these been added?

firebase-database/api/android/firebase-database.api
```diff
@@ -68,41 +69,17 @@ public final class dev/gitlive/firebase/database/FirebaseDatabase {
...
 	public final fun setValue (Lkotlinx/serialization/SerializationStrategy;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public final fun setValue (Lkotlinx/serialization/SerializationStrategy;Ljava/lang/Object;ZLkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun setValue$default (Ldev/gitlive/firebase/database/OnDisconnect;Lkotlinx/serialization/SerializationStrategy;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
 	public final fun updateChildren (Ljava/util/Map;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public final fun updateChildren (Ljava/util/Map;ZLkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun updateChildren$default (Ldev/gitlive/firebase/database/OnDisconnect;Ljava/util/Map;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
+	public final fun updateEncodedChildren (Ldev/gitlive/firebase/internal/EncodedObject;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
nbransby, line 82: why have these been added?

firebase-database/api/android/firebase-database.api
Daeda88: Because we have a bunch of inline methods that need access to the Native classes. Though those had been set as internal and annotated with @PublishedApi, this results in them still poping up in the API specification.

So it's essentially choosing between having the Native... classes in the api or these methods. I want for these because it makes the API "smaller"

firebase-database/api/android/firebase-database.api
Daeda88: See https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/555#discussion_r1655076296

firebase-database/api/android/firebase-database.api
Daeda88: To clarify: these methods are also marked as`@PublishedApi internal fun` and the access the native classes inside their code. As a result the Native classes no longer need to be Published, but the method does because public inline methods can only call publish/Published code

nbransby:
Understood, do these @PublishedApi function appear in the kdocs I wonder.

Does this change also mean the ios, android and is properties don't need to be extensions?

Daeda88:
Afaik internal methods marked with @PublishedApi are not accessible from kotlin, but they are from Java. So I don't think they need Kdoc. 

As for the extensions for android/ios/JS, they are still needed as they extend the classes that are in common code, e.g 'Query', to provide access to the Native classes (e.g. 'NativeQuery'). It used to be that these native classes, usually Typealiases, where public as well, but with this change all of that is internal. So in order to still support construction with platform specific code I added some 'invoke' methods to the companion objects though extension as well.

That means, you can still do 'Query(androidQuery)' on platform code, but no longer from common code and it requires an import of 'dev.gitlive.firebase.firestore.invoke'.

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -34,14 +34,16 @@ public expect val Firebase.firestore: FirebaseFirestore
 /** Returns the [FirebaseFirestore] instance of a given [FirebaseApp]. */
 public expect fun Firebase.firestore(app: FirebaseApp): FirebaseFirestore
 
-public expect class NativeFirebaseFirestore
+internal expect class NativeFirebaseFirestore
 
 public class FirebaseFirestore internal constructor(private val wrapper: NativeFirebaseFirestoreWrapper) {
```
Daeda88, line 39: Leaving comments to illustrate what I mean.

This class in common is the class users interact with. It is not an expect actual class and as such, to have the `android`/`ios`/`js` accessors, we need to write extension values on the platforms

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -34,14 +34,16 @@ public expect val Firebase.firestore: FirebaseFirestore
...
 
-    public constructor(native: NativeFirebaseFirestore) : this(NativeFirebaseFirestoreWrapper(native))
+    public companion object {}
+
+    internal constructor(native: NativeFirebaseFirestore) : this(NativeFirebaseFirestoreWrapper(native))
 
     // Important to leave this as a get property since on JS it is initialized lazily
-    public val native: NativeFirebaseFirestore get() = wrapper.native
+    internal val native: NativeFirebaseFirestore get() = wrapper.native
```
Daeda88, line 46: This used to be a common level alternative to `android/`ios`/`js` but it is now internal, because `NativeFirebaseFirestore` is now internal

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -34,14 +34,16 @@ public expect val Firebase.firestore: FirebaseFirestore
...
 
-public expect class NativeFirebaseFirestore
+internal expect class NativeFirebaseFirestore
 
 public class FirebaseFirestore internal constructor(private val wrapper: NativeFirebaseFirestoreWrapper) {
 
-    public constructor(native: NativeFirebaseFirestore) : this(NativeFirebaseFirestoreWrapper(native))
+    public companion object {}
+
+    internal constructor(native: NativeFirebaseFirestore) : this(NativeFirebaseFirestoreWrapper(native))
```
Daeda88, line 43: Since `NativeFirebaseFirestore` is now internal, the constuctor has to be as well

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -34,14 +34,16 @@ public expect val Firebase.firestore: FirebaseFirestore
 /** Returns the [FirebaseFirestore] instance of a given [FirebaseApp]. */
 public expect fun Firebase.firestore(app: FirebaseApp): FirebaseFirestore
 
-public expect class NativeFirebaseFirestore
+internal expect class NativeFirebaseFirestore
 
 public class FirebaseFirestore internal constructor(private val wrapper: NativeFirebaseFirestoreWrapper) {
 
-    public constructor(native: NativeFirebaseFirestore) : this(NativeFirebaseFirestoreWrapper(native))
+    public companion object {}
```
Daeda88, line 41: To circumvent https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/555/files#r1655371860 we introduce a companion object

firebase-firestore/src/commonMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -34,14 +34,16 @@ public expect val Firebase.firestore: FirebaseFirestore
 /** Returns the [FirebaseFirestore] instance of a given [FirebaseApp]. */
 public expect fun Firebase.firestore(app: FirebaseApp): FirebaseFirestore
 
-public expect class NativeFirebaseFirestore
+internal expect class NativeFirebaseFirestore
```
Daeda88, line 37: This is a typealias to `FIRFirestore`(ios) and `com.google.firebase.firestore.FirebaseFirestore`(android) or the wrapper on Javascript. To remove it from the public API, it is now internal but this has a cascading effect on
https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/555/files#r1655371860
https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/555/files#r1655371283

firebase-firestore/src/androidMain/kotlin/dev/gitlive/firebase/firestore/firestore.kt
```diff
@@ -39,9 +52,10 @@ public val LocalCacheSettings.android: com.google.firebase.firestore.LocalCacheS
     }
 }
 
-public actual typealias NativeFirebaseFirestore = com.google.firebase.firestore.FirebaseFirestore
+internal actual typealias NativeFirebaseFirestore = AndroidFirebaseFirestore
 
-public val FirebaseFirestore.android: NativeFirebaseFirestore get() = native
+public operator fun FirebaseFirestore.Companion.invoke(android: AndroidFirebaseFirestore): FirebaseFirestore = FirebaseFirestore(android)
```
Daeda88, line 57: With the public construcor gone (see https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/555/files#r1655371860), in order to still support `FirebaseFirestore(android: AndroidFirebaseFirestore)` we can add an extension to the companion object https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/555/files#r1655372358

This maintains the same syntax, but since it is an extension method, it does require an explicit import

nbransby:
ok so ill finish the breaking change of converting all .js, .android, .ios properties to extensions in https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/504

nbransby:
> Afaik internal methods marked with @PublishedApi are not accessible from kotlin, but they are from Java. So I don't think they need Kdoc.

Yes they dont need a docblock but I still wonder if dokka will publish them in the api docs

Issue #551, opened 2024-06-25
1.13.0 FirebaseFirestoreException: initializeFirestore() has already been called with different options


HansHolz09:
After upgrading to 1.13.0 I get the following error in Web/JS: 
`FirebaseFirestoreException: FAILED_PRECONDITION: initializeFirestore() has already been called with different options. To avoid this error, call initializeFirestore() with the same options as when it was originalIy called, or call getFirestore() to return the already initialized instance.`

nbransby:
Do you add a failing test for this? Or a minimal reproducer to this issue?

Daeda88:
Fixed in #562

PR #562, opened 2024-07-09
Fixed a crash for settings when iOS would access the same Firebase instance twice

### LINKED ISSUE #552: 1.13.0 Enabling/disabling Firestore network causes a crash on iOS
Looks like version 1.13.0 created a new issue where:
Firebase.firestore.enableNetwork()
or 
Firebase.firestore.disableNetwork()
would cause an app to crash:
`*** Terminating app due to uncaught exception 'FIRIllegalStateException', reason: 'Firestore instance has already been started and its settings can no longer be changed. You can only set settings before calling any other methods on a Firestore instance.'`

The reason seems to be that firestoreSettings gets recreated when this function is called.

This was not an issue in version 1.12.0.
The issue does not reproduce in Android app (1.13.0).

Daeda88:
Fixes #552

### DIFF
--- a/firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/internal/NativeFirebaseFirestoreWrapper.kt
+++ b/firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/internal/NativeFirebaseFirestoreWrapper.kt
@@ -10,6 +10,7 @@ import dev.gitlive.firebase.firestore.NativeWriteBatch
 import dev.gitlive.firebase.firestore.externals.clearIndexedDbPersistence
 import dev.gitlive.firebase.firestore.externals.connectFirestoreEmulator
 import dev.gitlive.firebase.firestore.externals.doc
+import dev.gitlive.firebase.firestore.externals.getFirestore
 import dev.gitlive.firebase.firestore.externals.initializeFirestore
 import dev.gitlive.firebase.firestore.externals.setLogLevel
 import dev.gitlive.firebase.firestore.externals.writeBatch
@@ -20,20 +21,33 @@ import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.await
 import kotlinx.coroutines.promise
 
+// There is currently no way to check whether Firestore was already initialized for a given app without actually initializing it
+// Therefore we keep track of this internally
+private val appsWithFirestore = mutableListOf<FirebaseApp>()
+
 internal actual class NativeFirebaseFirestoreWrapper internal constructor(
     private val createNative: NativeFirebaseFirestoreWrapper.() -> NativeFirebaseFirestore,
+    private val canUpdateSettings: () -> Boolean,
 ) {
 
-    internal actual constructor(native: NativeFirebaseFirestore) : this({ native })
+    internal actual constructor(native: NativeFirebaseFirestore) : this({ native }, { false })
     internal constructor(app: FirebaseApp) : this(
         {
             NativeFirebaseFirestore(
-                initializeFirestore(app, settings.js).also {
-                    emulatorSettings?.run {
-                        connectFirestoreEmulator(it, host, port)
+                if (appsWithFirestore.contains(app)) {
+                    getFirestore(app)
+                } else {
+                    initializeFirestore(app, settings.js).also {
+                        emulatorSettings?.run {
+                            connectFirestoreEmulator(it, host, port)
+                        }
+                        appsWithFirestore.add(app)
                     }
                 },
             )
+        },
+        {
+            !appsWithFirestore.contains(app)
         },
     )
 
@@ -41,7 +55,7 @@ internal actual class NativeFirebaseFirestoreWrapper internal constructor(
 
     actual var settings: FirebaseFirestoreSettings = FirebaseFirestoreSettings.Builder().build()
         set(value) {
-            if (lazyNative.isInitialized()) {
+            if (lazyNative.isInitialized() || !canUpdateSettings()) {
                 throw IllegalStateException("FirebaseFirestore has already been started and its settings can no longer be changed. You can only call setFirestoreSettings() before calling any other methods on a FirebaseFirestore object.")
             } else {
                 field = value
--- a/firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/internal/NativeFirebaseFirestoreWrapper.kt
+++ b/firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/internal/NativeFirebaseFirestoreWrapper.kt
@@ -23,15 +23,17 @@ import kotlinx.coroutines.promise
 
 // There is currently no way to check whether Firestore was already initialized for a given app without actually initializing it
 // Therefore we keep track of this internally
-private val appsWithFirestore = mutableListOf<FirebaseApp>()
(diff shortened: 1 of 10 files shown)


nbransby:
Why does accessing the same Firebase instance twice crash on iOS?

Daeda88:
iOS natively crashes when you change settings after doing something with firstore (e.g. disable the network). But the code contained this to sync settings:

```
    actual var settings: FirebaseFirestoreSettings = firestoreSettings { }.also {
        native.settings = it.ios
    }
```

So native settings would be set to `firestoreSettings { }` whenever a Kotlin FirebaseFirestore instance was created.
So if you do: 

```
Firebase.firestore.disableNetwork()
Firebase.firestore.enableNetwork()
```

What happens is:

- Grab default firestore app
- Set its settings to the default
- Disable network
- Grab default firestore app
- Set its settings to the default <-- This is not allowed

Something similar is happening on JS as per #551, Ill try and create a fix for that one too

Daeda88:
Latest commit adds a test & fixes #551

firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/internal/NativeFirebaseFirestoreWrapper.kt
```diff
@@ -20,28 +21,41 @@ import kotlinx.coroutines.GlobalScope
...
+        },
     )
 
     private data class EmulatorSettings(val host: String, val port: Int)
 
     actual var settings: FirebaseFirestoreSettings = FirebaseFirestoreSettings.Builder().build()
         set(value) {
-            if (lazyNative.isInitialized()) {
+            if (lazyNative.isInitialized() || !canUpdateSettings()) {
                 throw IllegalStateException("FirebaseFirestore has already been started and its settings can no longer be changed. You can only call setFirestoreSettings() before calling any other methods on a FirebaseFirestore object.")
```
nbransby, line 59: I thought we removed this check in favour of letting the native sdk throw the exception?

firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/internal/NativeFirebaseFirestoreWrapper.kt
```diff
@@ -20,28 +21,41 @@ import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.await
 import kotlinx.coroutines.promise
 
+// There is currently no way to check whether Firestore was already initialized for a given app without actually initializing it
+// Therefore we keep track of this internally
+private val appsWithFirestore = mutableListOf<FirebaseApp>()
+
```
nbransby, line 27: what are we doing exactly that is straightforward on android but  not supported natively on ios and js? The other option here would be to cater for the lowest common denominator, in other words don't support on android either,  but I still not exactly clear what the differences are?

Daeda88: So there are two things:

First and foremost, the biggest difference between the mobile platforms and JS is that for the mobile platforms, the FirebaseFirestore object internally keeps a state machine that allows it to change settings until it is first used. So, you can do:

```
firestore.settings.host = "1.1.1.1"
firestore.enableNetwork()
firstore.settings.host = "2.2.2.2" // This is not allowed
```

Javascript on the other hand has no such stateMachine with regards to settings. You can either get a FirebaseFirestore instance with whatever settings it was previously initialized, or you can initialize it with settings provided it has not been initialized before (there's no telling whether it is btw). The only one you can do is enable the emulator, which you can do after initialization but only before doing any operations.

This difference is the main reason for this major discrepancy. The approach of this library has always been to get the FirebaseFirestore object and change its settings before operations are executed. If you look at the `setSettings` method on JS pre 1.13.0, you'll realize that it would've crashed if you called the method as it calls the initializationCode after it has already gotten an initialized instance.

I guess it __might__ be possible to essentially remove the state machine logic out of the library, so that you must provide the settings when getting a FirebaseFirestore instance, but this might have unknown side effects and is a breaking change for sure.

The second problem is smaller, and could be solved by omission: Settings on Android are fully readable, on iOS readable except for the cache settings, and on JS (due to initialization mentioned above) not readable at all. As of this PR, we get the settings on Android/iOS (though we miss a few cache settings on iOS) and we track it ourselves for JS.

firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/internal/NativeFirebaseFirestoreWrapper.kt
Daeda88: The thing is that there is no native sdk to fall back to, as explained in the other remark. We shouldnt be calling initializeFirestore here yet, as settings may still be updated after this.

firebase-firestore/src/jsMain/kotlin/dev/gitlive/firebase/firestore/internal/NativeFirebaseFirestoreWrapper.kt
nbransby: Ok well for the second issue I think removing the getter so you can't read settings on any platform is sensible.

As for the first issue, looking at https://firebase.google.com/docs/firestore/manage-data/enable-offline#configure_offline_persistence, I suggest the following: 
-  make it only possible to set all the settings at once so `firestore.settings.host = "1.1.1.1` shouldn't be possible (which will be the case once the getter is removed)
- on JS, since `Firebase.firestore.settings = firestoreSettings { ... }` needs to call `initializeFirestore` before `getFirestore` is called can the js property be lazy so `getFirestore` is not called doing  `Firebase.firestore.settings = firestoreSettings { ... }`
- if the user attempts to set the settings after doing another operation which means `getFirestore` has already been called then let the js sdk (and others) throw the relevant error.

Daeda88: Agreed:

- Removed settings property in favour of `setSettings(builder)` and `applySettings(settings)` methods. This is an api change compared to 1.13.0 but I think we're both fine with that.
- The JS implementation now triggers "setting" the settings if the firestore object has been initialized. This should ensure you get a native crash if you try and set settings after already having initialized the firestore object.

nbransby: Lets keep the syntax matching the [android sdk](https://github.com/firebase/snippets-android/blob/bdab1d019d41c1e034ecc1d63d30c7071af3d79d/firestore/app/src/main/java/com/google/example/firestore/kotlin/DocSnippets.kt#L112-L118) with a [setter only property](https://stackoverflow.com/a/54983086/737857) rather that two new functions

Daeda88: Its not an officially supported feature, so it has some downsides as your link shows. Regardless, Ive implemented it as requested

Tweener:
Hi 👋 
I'm also having the same issue on iOS. 
Do you guys have an idea when a new release can be expected?
Thanks!

Btw, love the work you guys have been doing with this library! 🫶

PR #592, opened 2024-08-27
Updated firebase-cocoapods to "11.1.0"

siarhei-luskanau:


Daeda88 (approved):

siarhei-luskanau:
@Daeda88 Hi, I see you have approved the PR. 
This PR contains the compilation error in `firebase-auth/src/iosMain/kotlin/dev/gitlive/firebase/auth/auth.kt` file. I can't resolve some imports

Daeda88:
Ill try and have a look to see if I can understand why

Daeda88:
@siarhei-luskanau I checked out your branch. The reason its failing is because the FIRAuthErrorCode enum was moved to Swift and Kotlin Native is not able to resolve it. I think the only way to solve it is by replacing the enum codes with the actual values https://github.com/firebase/firebase-ios-sdk/blob/2f6ac4c2c61cd57c7ea727009e187b7e1163d613/FirebaseAuth/Sources/Swift/Utilities/AuthErrors.swift#L51

I recall someone ran into a similar issue for another module, but I cannot find it. Maybe @nbransby remembers.

nbransby (approved):

PR #594, opened 2024-08-27
Revert to API 21 as minSDK for all but Auth

### LINKED ISSUE #591: Lower Android minSdk
Hi,

I noticed that the library's min SDK version was raised to 23 in this [commit](https://github.com/GitLiveApp/firebase-kotlin-sdk/commit/267bfd14ae1af8ef1f033ae90cbb5cbd7a2bd393).

It seems like this change might have been made by mistake. The commit message mentions Android 14's restriction on apps with a `targetSdk` lower than 23. However, this restriction only applies to the `targetSdk` of the app, not the `minSdk` of the library.

Wouldn't it make more sense to revert the min SDK version change as it was (at 19) and let the application itself decide on the appropriate target SDK? Raising the `minSdk` to 23 unnecessarily limit the library's compatibility with apps that support older Android versions.

Daeda88:
Fixes #591

### DIFF
--- a/firebase-auth/src/commonTest/kotlin/dev/gitlive/firebase/auth/auth.kt
+++ b/firebase-auth/src/commonTest/kotlin/dev/gitlive/firebase/auth/auth.kt
@@ -71,6 +71,7 @@ class FirebaseAuthTest {
     }
 
     @Test
+    @Ignore
     fun testFetchSignInMethods() = runTest {
         val email = "test+${Random.nextInt(100000)}@test.com"
         var signInMethodResult = auth.fetchSignInMethodsForEmail(email)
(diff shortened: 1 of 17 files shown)


firebase-auth/src/commonTest/kotlin/dev/gitlive/firebase/auth/auth.kt
```diff
@@ -71,6 +71,7 @@ class FirebaseAuthTest {
     }
 
     @Test
+    @Ignore
```
Daeda88, line 74: Could also just delete this test. It's gonna fail cause the methods it tests are deprecated and only return an empty list

Issue #604, opened 2024-08-29
Encoder is no longer accessible?


mr-kew:
I was using the firebase encoder to encode a single values to be used in where statements. It was not flawless because the encoder ignored my `@Serializable(with = ...)` for some reason, but it worked really well.

But now the `encode` & `firebaseSerializer` methods have been made internal, so I can't use them anymore. And I struggle to find any alternatives. Why is this functionality internal all of the sudden? Does this mean I am just screwed now?

### My use case:

```
import dev.gitlive.firebase.FirebaseEncoder
import dev.gitlive.firebase.encode
import dev.gitlive.firebase.firebaseSerializer
... // These imports no longer work, even the @file:Suppress("INVISIBLE_MEMBER", "INVISIBlE_REFERENCE") hack does not help

internal object PrimitiveFirebaseEncoder {
    inline fun <reified T> encodeToAny(value: T): Any? = value?.let {
        encode(it.firebasePrimitiveSerializer(), it) {
            encodeDefaults = true
        }
    }

    /**
     * Allows to override/specify default serializer for a standalone primitive value serialization using [FirebaseEncoder].
     */
    @Suppress("UNCHECKED_CAST")
    private inline fun <reified T : Any> T.firebasePrimitiveSerializer() = when (this) {
        is Instant -> InstantTimestampSerializer
        is LocalDate -> LocalDateTimestampSerializer
        is LocalTime -> LocalTimeIntSerializer
        is LocalDateTime -> LocalDateTimeTimestampSerializer
        is Duration -> DurationSecondsSerializer
        is DayOfWeek -> DayOfWeekIntSerializer
        else -> this.firebaseSerializer()
    } as KSerializer<T>
}
```
**Usage:** 
```
val time: Instant = ...
query
    .where { "time" greaterThan PrimitiveFirebaseEncoder.encodeToAny(time) }
    ...
```

Daeda88:
It was indeed removed from the public facing api. I guess we never considered adding encoding support for where clauses.

My proposal would be to add support for it to the framework.

Of course, that doesnt solve oyur issue yet. There's two things you can do right now to proceed.
1. Just serialize yourself. By that I mean instead of doing

```kotlin
query
    .where { "time" greaterThan PrimitiveFirebaseEncoder.encodeToAny(time) }
```

do

```kotlin
query
    .where { "time" greaterThan time.asPrimitive }
```

where `asPrimitive` is used by both the serializer and the where clause.

[...]

Issue #590, opened 2024-08-26
Facing issue while trying to read collection from firestore


ahmad244221:
I am facing ClassNotFoundException while reading collection.  I have added all the needed dependencies in gradle file. I did the signin process with custom token.

`Caused by: java.lang.ClassNotFoundException: Didn't find class "dev.gitlive.firebase.DecodersKt"`

Below is the shared module gradle dependencies

`val commonMain by getting {
            dependencies {
               implementation(libs.kotlinx.coroutines.core)
                implementation(libs.ktor.client.core)
                implementation(libs.ktor.client.content.negotiation)
                implementation(libs.ktor.serialization.kotlinx.json)
                implementation(libs.ktor.client.logging)
                implementation(libs.koin.core)
                implementation(libs.sql.coroutines.extensions)
                implementation(libs.firebase.firestore)
                implementation(libs.firebase.common)
                implementation(libs.kotlinx.serialization.json)
                implementation(libs.firebase.auth)
            }
        }`

These are android module denpencies.

`dependencies {
    implementation(projects.shared)
    implementation(libs.compose.ui)
    implementation(libs.compose.ui.tooling.preview)
    implementation(libs.compose.material3)
    implementation(libs.androidx.activity.compose)
    implementation(libs.androidx.appcompat)
    implementation(libs.androidx.constraintlayout)
    implementation(libs.material)
    implementation(libs.androidx.lifecycle.livedata.ktx)
    implementation(libs.androidx.lifecycle.viewmodel.ktx)
    implementation(libs.androidx.navigation.fragment.ktx)
    implementation(libs.androidx.navigation.ui.ktx)
    debugImplementation(libs.compose.ui.tooling)
    implementation(libs.koin.android)
    implementation(platform(libs.google.firebase.bom))
    implementation(libs.firebase.common.ktx)
    implementation(libs.google.firebase.common)
}`

Please suggest if anyone faced this issue or some workaround to fix this issue.

nbransby:
@Daeda88 any clue on this?

ahmad244221:
Below are lines of code I am trying to read collection after calling signInWithCustomToken() - 

`1.  val tokenResponse = auth.signInWithCustomToken(customtoken)
      val response = firestore.collection("collectionPath")
                .document("documentPath").get().data<MyDataClass>(MyDataClass.serializer())
            println("response... $response")
			

2. val tokenResponse = auth.signInWithCustomToken(customtoken)
[...]

PR #605, opened 2024-08-29
Moves ValueWithSerializer back into public API

Daeda88:
It had become part of commons-internal which kind of defeats the purpose.

Also made FirebaseEncoder/FirebaseDecoder an interface in the public API. Useful for writing custom Serializers that have custom behaviour on Firebase


nbransby:
> It had become part of commons-internal which kind of defeats the purpose.

I'm going to need more of an explanation, is it to fix this issue https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/590 ?

Daeda88:
No I actually ran into it myself. The ValueWithSerializer class was originally in the regular commons library and allows encoding a value when dealing with File paths. But it's no longer accessible so I moved it back. Though I'm not entirely sure if this was introduced by Splendo or GitLive. Technically you can achieve the same behavior with a custom SerializerModule though, this does add some convenience

nbransby:
> allows encoding a value when dealing with File paths

Could you give a code example of what you mean? A search of ValueWithSerializer in the README doesn't return any results.

mr-kew:
@Daeda88 Does this somehow relate to [#604](https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/604)?

Daeda88:
Dropping in favour of #607

PR #620, opened 2024-09-05
FIX target OAuth provider not accessable

DatL4g:
It's not possible to access the target specific OAuth provider, as the field was marked internal and a extension field with the same name was present, resulting in an error like this: `Cannot access 'val android: OAuthProvider': it is internal in 'dev/gitlive/firebase/auth/OAuthProvider'.`

### DIFF
--- a/firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
+++ b/firebase-auth/src/androidMain/kotlin/dev/gitlive/firebase/auth/credentials.kt
@@ -52,7 +52,7 @@ public actual object GoogleAuthProvider {
     }
 }
 
-public actual class OAuthProvider(val android: AndroidOAuthProvider) {
+public actual class OAuthProvider(public val android: AndroidOAuthProvider) {
 
     public actual constructor(
         provider: String,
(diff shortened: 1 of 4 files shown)


nbransby:
Why can't you access the extension field with the same name? That's public is it nofm

DatL4g:
The IDE and compiler will always try to access the class properties and functions before the extension functions.
So they don't even know that a public equivalent exists.
And since the extension is located in the same package it's not even possible to change the import statement.

nbransby:
That seems foolish on the complier's part considering the other property is not publicly accessible

Reedyuk:
I think this was broken in a previous release (2.0.0) but its fixed in 2.1.0.

Here I am using the android extension fine and even outputs to console.

Uploading Screenshot 2024-
![Screenshot 2024-09-09 at 09 42 54
09-09 at 09.50.31.png…]()

DatL4g:
Another workaround would be to let the constructor property be internal but rename it to `_android` for example.
What do you think of that change @nbransby

DatL4g:
> That seems foolish on the complier's part considering the other property is not publicly accessible 

I don't think that's a problem on the Kotlin compiler part btw, since extension functions are meant to extend objects and I guess the compiler will simply ignore the extension if the name, return type and package is exactly the same as the one already existing.
So the code will just not be compiled.

DatL4g:
@Reedyuk thanks for the hint, I was contributing to another library building on top of this, and yes it seems to use `2.0.0`
I will update the version and give some feedback then :)

Reedyuk:
I'm going to close this PR but if you still have the issue, then re-open.

Issue #520, opened 2024-06-02
Crashlytics - iOS - setCustomKey - wrong parameters order


alaegin:
Hello!

I started using setCustomKey in the Crashlytics library. 
Unfortunately, the current implementation has some serious issue: the order of parameters in setCustomKey is wrong.

Here is the code from my app:
```kotlin
.distinctUntilChanged()
.onEach { balance ->
    logger.i("Setting balance in Crashlytics: $balance")
    Firebase.crashlytics.setCustomKey("balance", balance)
}
```

Here is the result in Crashlytics report:
![image](https://github.com/GitLiveApp/firebase-kotlin-sdk/assets/10329086/ed17b509-b52d-4094-ad5b-9df77ff08216)

Here is the [Firebase](https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=ios#add-keys) documentation:
![image](https://github.com/GitLiveApp/firebase-kotlin-sdk/assets/10329086/dff5a2ea-64d1-4c12-9e2b-961644048a14)
 
It seems like the current implementation was done wrong because of the difference between Android (key, value) and iOS (value, key) Firebase libraries.

nbransby:
Sounds like you know how to fix it! Any chance of a PR?

alaegin:
@nbransby Hah, I'll take a look

kaidotarma:
Here is the fix PR: https://github.com/GitLiveApp/firebase-kotlin-sdk/pull/646

PR #646, opened 2024-10-13
Fix iOS setCustomKey wrong param order

kaidotarma:
Fixes https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/520

According to the docs: https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=ios#add-keys
the order of setCustomValue parameters is: value, key

setCustomKeysAndValues still expects to order of the map elements to be key, value - so no changes there.


nbransby:
nice work @kaidotarma !

PR #652, opened 2024-10-17
Bump firebase-bom to 33.5.0

fethij:


nbransby:
Thanks for this, and sorry it sat so long. It's obsolete now — master is on firebase-bom 33.15.0, and #777 covers the bump to 34.4.0.

Issue #699, opened 2025-01-25
Add [Query].[count()] to [FireStore] for [Android/IOS]


baseemomar:
Add support for the` count() `function in Query to retrieve the count of documents matching a query without fetching the entire data. This will return an AggregateQuery which can be executed using `get(AggregateSource)`. This feature significantly reduces network usage and Firestore read costs, especially for large datasets.

chris-hatton:
This is also the only functionality that's keeping me from removing a direct KMP Cocoapods dependency on Firebase, which really slows down Kotlin builds.  `count()` will be a welcome addition to the wrapping.

chris-hatton:
Duplicate of #657

anggrayudi:
Any updates on this issue?

anggrayudi:
This has shipped. `Query.count()` (along with `sum()` and `average()`) was merged via #830 and is available in **v2.5.0** on Maven Central:

```kotlin
val total: Long = firestore.collection("users").where { ... }.count()
```

Implementations are present for Android, Apple and JS.

Editing my earlier note for accuracy: the merged API differs in *shape* from the original request — it's a `suspend fun count(): Long` rather than an `AggregateQuery` you execute via `get(AggregateSource)` — but nothing is functionally missing. `AggregateSource` has exactly one constant, `SERVER` (checked against `firebase-firestore` 26.1.2), and the Apple/JS equivalents are likewise server-only, so there is no other source that could be passed.

This one looks closeable.

nbransby:
There's an open PR for this: #785 by @Daeda88, which adds aggregate query support to Firestore — `AggregateQuery`, `AggregateField.Count` and `Query.count()` — across Android, Apple and JS, with tests.

nbransby:
Closing: `Query.count()` is on master, shipped in #830 and available from 2.5.0, along with `sum()` and `average()`.

```kotlin
val total: Long = firestore.collection("users").where { ... }.count()
```

It is a suspend function that always queries the server, which is the only aggregate source the native SDKs offer. See #657 for the same request and #785 for the proposed `AggregateQuery` shape.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #657, opened 2024-10-24
Add Query.count to FireStore for Android, iOS


mrHerintsoaHasina:
Hi

I found missing API for aggregation queries as documented [here](https://firebase.google.com/docs/firestore/query-data/aggregation-queries#kotlin+ktx). In my case, I want to use the `count` method. But it would be nice also to include the others methods like `sum` and `average`.

Feinate:
is there any news for this, it would be very usefull !

chris-hatton:
Duplicate of #699

anggrayudi:
Any update on this?

anggrayudi:
This has shipped. `count()`, `sum()` and `average()` were all merged via #830 and are available in **v2.5.0** on Maven Central:

```kotlin
val total: Long = query.count()
val revenue: Double = query.sum("amount")
val avg: Double? = query.average("amount")
```

Implementations are present for Android, Apple and JS.

Editing my earlier note for accuracy: these are plain `suspend` functions with the aggregate source fixed to SERVER, rather than an `AggregateQuery` executed via `get(AggregateSource)`. That is a shape difference only — `AggregateSource` has exactly one constant, `SERVER` (checked against `firebase-firestore` 26.1.2), and the Apple/JS equivalents are server-only too, so no alternative source exists to expose.

This one looks closeable.

nbransby:
There's an open PR for this: #785 by @Daeda88, which adds `AggregateQuery`, `AggregateField.Count` and `Query.count()` across Android, Apple and JS, with tests.

Note this duplicates #699 — worth consolidating once #785 lands.

nbransby:
Closing: this is on master. `Query.count()`, `sum()` and `average()` shipped in #830 and are in 2.5.0 and later.

```kotlin
val total: Long = query.count()
val revenue: Double = query.sum("amount")
val avg: Double? = query.average("amount")
```

They are plain suspend functions against the server rather than an `AggregateQuery` you execute yourself. #785 proposes the `AggregateQuery` shape alongside a bigger SDK bump, and can carry that discussion if there is still appetite for it.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #702, opened 2025-02-05
trying to implement count function

beardykun:


chris-hatton:
@beardykun Mind saying why you closed this?  It looks like the solution to #657 and #699 - did you identify an issue with it?

Issue #689, opened 2024-12-22
[Common] [Storage] Can't upload data from common module


Monabr:
Hi. I am trying to upload my bite array to Cloud Storage.

I am doing this `.putBytes(Data(byteArray))` but I see that expected declaration have no constructor. 

What should I do to upload data from common module?

Ashkan-san:
did you find a solution to this? i also dont understand how to upload files

pontomedon:
Let's hope #704 gets merged soon, then it will work just fine even on desktop :)

pontomedon:
@nbransby thanks for merging #704 - I think this issue can be closed now too.

Issue #533, opened 2024-06-16
Add Firebase Storage for JVM target


CRamsan:
I need firebase storage for all platforms but I am starting with JVM for now.

The work is starting on the java sdk here: https://github.com/GitLiveApp/firebase-java-sdk/issues/4

pontomedon:
Let's hope #704 gets merged soon, then it will work just fine even on desktop :)

PR #704, opened 2025-02-18
Adding JVM support for Firebase Storage

pontomedon:
This is based on previous work by Cesar Ramirez and it is obviously depending on work done on the java sdk: https://github.com/GitLiveApp/firebase-java-sdk/pull/44

I know the checks are failing, I'll need some input on how I can publish a snapshot of `firebase-java-sdk` to make them succeed. It all works locally when deploying to maven local.

I needed to make sure we never call the firebase storage API with `null` metadata - when doing this on a desktop target, the `CONTENT_LENGTH = 0` header gets removed somehow, causing the firebase backend to respond with a HTTP 411. Passing defaults causes a few bytes to be serialized in the first requrest before the upload, which circumvents the problem.

The root cause for this must be in the http driver somewhere, as it doesn't happen when running the same code on Android.


pontomedon:
fixes #689 and #533

pontomedon:
You'll have to do the following:
* check out https://github.com/GitLiveApp/firebase-java-sdk/pull/44
* update `gradle.properties` and set the version `0.4.9-AbelVolpi-SNAPSHOT`
* deploy `firebase-java-sdk` to your maven local `./gradlew publishToMavenLocal`
* check out this branch
* update `libs.versions.toml` and set `gitlive-firebase-java-sdk = "0.4.9-AbelVolpi-SNAPSHOT"`
* update `gradle.properties` and set all `firebase-*.version` to `2.1.1-AbelVolpi-SNAPSHOT`
* deploy `firebase-kotlin-sdk` to your maven local `./gradlew publishToMavenLocal`
* in your project, make sure you have `mavenLocal()` in your repository list and depend on `firebase-kotlin-sdk` in version `2.1.1-AbelVolpi-SNAPSHOT`

alternatively: you can open ~/.m2/repository and copy the whole `dev/gitlive` folder into your repository, under a subfolder `.localMaven`, and add this directory as a local maven to your repositories: `maven { url = URI("file://${rootProject.projectDir}/.localMaven") }`, that way it will work on CI as well.

Issue #715, opened 2025-04-07
Add missing implementations for iOS/Android on performance


blackwiz4rd:
Hello,
I'm willing to add the missing implementations for Firebase Performance SDK for iOS/Android.
So I will try to make a PR as soon as possible for this feature.
I hope it will be helpful.
Regards!

blackwiz4rd:
I'm almost done with the implementation.
Missing tasks:
- iOS implementation for `HttpMetric`
- Run tests on Android

If you guys want to see my live implementation checkout https://www.youtube.com/watch?v=kR--EWdbHng, it might inspire you to contribute to this project as well :)

nbransby:
Linking your PR for visibility: #726.

Note that master has moved since — the `firebase-perf` module now has `Trace` and `performance` for Android, Apple, JS and JVM, including the trace attribute APIs. The part of #726 not yet on master is `HttpMetric`, so a rebase down to that would likely be the quickest path in.

PR #726, opened 2025-05-27
Added missing Firebase Performance SDK implementation for Android and iOS

blackwiz4rd:
Implementation:
- full tests coverage for Android
- implemented more fields for Android
- implemented iOS Firebase Performance

Reference issue: https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/715

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+github: [GitLiveApp]


firebase-perf/src/iosMain/kotlin/dev/gitlive/firebase/perf/metrics/Trace.kt
```diff
@@ -23,4 +23,20 @@ public actual class Trace internal constructor(internal val ios: FIRTrace?) {
     public actual fun putMetric(metricName: String, value: Long) {
         ios?.setIntValue(value, metricName)
     }
+
+    public actual fun getAttribute(attribute: String): String? {
+        TODO("Not yet implemented")
```
blackwiz4rd, line 28: check here

PR #784, opened 2025-11-29
Add FirebaseCrashlytics.recordException overload that takes key-value pairs

ARTI1208:
Added support for recordException overload that takes customKeys


ARTI1208:
@nbransby hi, could you review please?

nbransby (approved):

Issue #786, opened 2025-12-29
Crash with FirebaseAuth.sendSignInLinkToEmail on iOS


treitter:
Hello, I'm getting the following crash on iOS:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[FIRActionCodeSettings setDynamicLinkDomain:]: unrecognized selector sent to instance 0x600000012250'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001804f39e8 __exceptionPreprocess + 172
	1   libobjc.A.dylib                     0x000000018009c084 objc_exception_throw + 72
	2   CoreFoundation                      0x00000001805092a8 +[NSObject(NSObject) instanceMethodSignatureForSelector:] + 0
	3   CoreFoundation                      0x00000001804f7cb8 ___forwarding___ + 1196
	4   CoreFoundation                      0x00000001804fa1fc _CF_forwarding_prep_0 + 92
	5   DoubleStrain dev.debug.dylib        0x0000000105d2be78 kfun:dev.gitlive.firebase.auth#toIos__at__dev.gitlive.firebase.auth.ActionCodeSettings(){}cocoapods.FirebaseAuth.FIRActionCodeSettings + 1472
	6   DoubleStrain dev.debug.dylib        0x0000000105d27a28 kfun:dev.gitlive.firebase.auth.FirebaseAuth#sendSignInLinkToEmail#suspend(kotlin.String;dev.gitlive.firebase.auth.ActionCodeSettings;kotlin.coroutines.Continuation<kotlin.Unit>){}kotlin.Any + 608	7   DoubleStrain dev.debug.dylib        0x00000001055b95c4
```

It seems to be caused by calling `setDynamicLinkDomain` in `firebase-ios-sdk` which doesn't exist in the version my app depends upon (`12.4.0` but also if I upgrade to `12.7.0`). Dynamic domains are deprecated and about to go away entirely so losing access to this upstream API wouldn't be a loss at all.

I can work around this locally by swizzling `setDynamicLinkDomain` to a no-op but it would be much cleaner to not have to do that and it's not a best practice.

I think this could be fixed by upgrading the `firebase-cocoapods` dependency to a newer version. I'm not sure exactly which one but it seems to be at least one major version behind.

umbrellait-Zafar-Abdulloev:
Facing the same issue

anggrayudi:
[...]

PR #813, opened 2026-04-09
Optionally set the dynamic link domain

robholmes:
Dynamic link domains are now deprecated and if the firebase iOS SDK version is 12+ then the function has been removed. Causing the following crash, even though the `dynamicLinkDomain` is `null`. This at least supports not calling the removed function if the value is null.

```
Fatal Exception: NSInvalidArgumentException -[FIRActionCodeSettings setDynamicLinkDomain:]: unrecognized selector sent to instance
```

## Why?

Due to a bug in iOS 26.4 and the Firebase iOS SDK (detailed here: https://github.com/firebase/firebase-ios-sdk/issues/16033), a fix was applied in 12.12.0 and above of the Firebase iOS SDK.

So to fix the crash, we need to update the Firebase iOS SDK version to 12.12.0 and above in xcode.

But doing this then leads to the removed `setDynamicLinkDomain` crash above.


robholmes:
This should also fix: https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/786

marcorighini (approved):

marcorighini:
@Reedyuk do you think this can be merged and a new version be released? Thank you

nbransby (approved):

PR #825, opened 2026-05-28
Fix/realtime database purge outstanding writes

### LINKED ISSUE #821: [Firebase Database] purgeOutstandingWrites missing from Common API and native .ios is internal
Description: On Android, purgeOutstandingWrites() can be accessed via extensions, but on iOS, it is currently impossible to implement because:

1. It is not part of the commonMain API.
2. The FirebaseDatabase.ios property is marked internal, so we cannot call the native FIRDatabase method in our own iosMain source sets.

Requested Action: Please expose purgeOutstandingWrites() in the common API, or change the visibility of the native .ios property to public.

michaelprichardson:
FIxes #821

### DIFF
--- a/firebase-database/src/jsMain/kotlin/dev/gitlive/firebase/database/database.kt
+++ b/firebase-database/src/jsMain/kotlin/dev/gitlive/firebase/database/database.kt
@@ -85,6 +85,8 @@ public actual class FirebaseDatabase internal constructor(internal val js: Datab
     public actual fun goOffline(): Unit = rethrow { jsGoOffline(js) }
 
     public actual fun goOnline(): Unit = rethrow { jsGoOnline(js) }
+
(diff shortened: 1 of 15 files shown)


firebase-database/src/jsMain/kotlin/dev/gitlive/firebase/database/database.kt
```diff
@@ -85,6 +85,8 @@ public actual class FirebaseDatabase internal constructor(internal val js: Datab
     public actual fun goOffline(): Unit = rethrow { jsGoOffline(js) }
 
     public actual fun goOnline(): Unit = rethrow { jsGoOnline(js) }
+
+    public actual fun purgeOutstandingWrites(): Unit = throw UnsupportedOperationException("purgeOutstandingWrites is not supported by the Firebase Realtime Database JS SDK.")
```
nbransby: probably best this is a no-op or log a warning

nbransby (approved):

Issue #835, opened 2026-07-05
Update iOS cinterop binding from firebase-ios-sdk 11.8.0 to a current 12.x


CohesionBrew:
### Summary

The iOS cinterop bindings are generated against `firebase-cocoapods = "11.8.0"` (in `gradle/libs.versions.toml`). That value is unchanged on `master` and in the latest release **v2.5.0 (2026-05-21)**, while `firebase/firebase-ios-sdk` is now at **12.15.0 (2026-06-16)** — the entire 12.x line has shipped since 11.8.0 (released early 2025).

Because the binding is exact rather than a range, consumers must pin their Podfile's `Firebase/*` pods to `11.8.0` to match. Pairing the bindings with newer pods (we tested 12.10.0) produces runtime `NSException`s, so the pin is a hard requirement, not a floor. The practical effect is that apps depending on this SDK are frozen ~18 months behind on the native Firestore/Auth/gRPC-C++/BoringSSL stack and can't pick up the security and bug fixes shipped across firebase-ios 12.x.

### Ask

Could the iOS cinterop binding be regenerated against a current firebase-ios-sdk 12.x (ideally the latest 12.15.0) and released? Even a single 12.x-pinned release would let downstream apps move their Podfiles forward in lockstep.

### Context / offer to help

- Consumer Podfile currently must pin `Firebase/* '11.8.0'` to match the 2.4.0 / 2.5.0 cinterop.
- Firebase **Android** BOM already tracks current (33.x on master), so this is specifically the iOS CocoaPods binding that's behind.
- Happy to test a pre-release against a real Compose Multiplatform app (anonymous / Apple / Google auth + Firestore one-shot & realtime + AppCheck + RemoteConfig) if that would help validate a bump.

Thanks for maintaining this — the KMP-everywhere Firestore story is exactly what we needed.

herrbert74:
Thanks for tracking this — we’re hitting the same constraint from another angle and would really appreciate a 12.x cinterop bump when you can prioritize it.

We’re on a Compose Multiplatform app that uses:

dev.gitlive:firebase-auth **2.5.0** (cinterop still built against Firebase iOS **11.8.0**)
[KMPNotifier](https://github.com/mirzemehdi/KMPNotifier) **2.0**, which declares SwiftPM firebase-ios-sdk exact 12.14.0 for Messaging
[...]

PR #836, opened 2026-07-12
Feat/cocoapods to swiftpm

nbransby:
migrate from CocoaPods to SwiftPM for the firebase-ios-sdk dependency


documentation/ios-firebase-linking.md
```diff
@@ -0,0 +1,116 @@
+# Linking the Firebase iOS SDK (SwiftPM)
+
+The Firebase Kotlin SDK is distributed as **Maven Central klibs** and consumed by a
+Kotlin Multiplatform (KMP) app, whose iOS framework is then added to an Xcode project. On iOS the
+GitLive klibs call the Firebase Obj-C API through cinterop, but — as before — **the official
+Firebase iOS SDK is not a transitive dependency**: it is not bundled in the klibs and does not flow
+across the Maven boundary. The **downstream KMP app** is responsible for making `firebase-ios-sdk`
+available when it links its own iOS framework.
+
+This is unchanged in spirit from the old CocoaPods setup ("Firebase iOS frameworks must be linked
```
timofeys1: Hey! We actually bundle SwiftPM dependencies metadata in the KMP publication metadata. It will be in the root component in the `-swiftpm-metadata.json` file referenced by the `swiftPMDependenciesMetadataElements` variant. On the library consumer's side, we unwrap this metadata to produce machine code for transitive SwiftPM dependencies, and then we also require wiring the Xcode project with the Gradle-generated package containing all transitive SwiftPM dependencies.

If this mechanism doesn't work in your case, I would be very curious to take a look at the reproducer.

nbransby: Hey thanks @timofeys1, we'll look into that! Are the details on that documented anywhere yet?

nbransby: @timofeys1 could you check the latest changes in the docs now correctly reflect the current state of SPM support? Thanks!

timofeys1: We don't have a document that describes exactly how transitive SwiftPM dependencies work, but we mention [here](https://kotlinlang.org/docs/multiplatform/multiplatform-spm-import.html) that transitive SwiftPM dependencies work transparently for the consuming project.

Everything else lgtm. The remote case will be supported in 2.4.20, so feel free to update the document when this version is released.

Thanks for working on this migration! If you discover something that doesn't work (as expected), then we will be happy to chat and help in our [Kotlinlang Slack](https://kotlinlang.slack.com/archives/C09TW68099C) or in [YouTrack](https://youtrack.jetbrains.com/issues/KT).

hohteri:
@nbransby any chance we get this merged sometime soon?

nbransby:
Pushed two commits that also close #835 — bumping `firebase-ios-sdk` from 11.8.0 to **12.17.0**.

**This grows the scope of the PR after @timofeys1's review**, so it's worth a re-read.

### Breaking change

Minimum apple deployment targets rise to **iOS 15 / tvOS 15** (from 13). macOS stays 10.15. Forced by firebase-ios-sdk 12.0.0, which declares `platforms: [.iOS(.v15), .macCatalyst(.v15), .macOS(.v10_15), .tvOS(.v15), .watchOS(.v7)]`.

### Two apple-only behaviour changes

Both are API removals in firebase-ios-sdk 12.0.0, verified by diffing the upstream tags:

- `FIROptions.trackingID` removed (long deprecated as unused) → `FirebaseOptions.gaTrackingId` is no longer round-tripped on apple. Android/JVM/JS unaffected.
- `FIRActionCodeSettings.dynamicLinkDomain` removed with the dynamic links shutdown → the apple binding now sets only `linkDomain`, which 11.8 introduced as the replacement. `ActionCodeSettings.dynamicLinkDomain` still works on android and jvm.

Nothing else in the bound Obj-C surface changed: every product and clang module the build declares still exists in 12.x, so **no module build file needed editing** — all 12 `swiftPMDependencies` blocks read the version and deployment targets from the catalog.

### Why this fixes #835 beyond the version number

`version = from(...)` is a lower bound, so the requirement becomes `>=12.17.0, <13.0.0`. That resolves the graph deadlock @herrbert74 described — a `<12` requirement could not co-resolve with KMPNotifier's `exact 12.14.0`.

### Verified locally (Xcode 26.2 / Swift 6.2.3)

- SwiftPM resolves to 12.17.0 (`Package.resolved` confirmed)
- `compileKotlinIosSimulatorArm64` green across all 12 modules
- `compileKotlinTvosSimulatorArm64` and `compileKotlinMacosArm64` green — worth calling out because `pull_request.yml` has no macOS or tvOS jobs, so **CI cannot catch a regression on the platform whose floor is actually moving**
- `linkDebugTestIosSimulatorArm64` green for firebase-app and firebase-auth
- `lintKotlin` and the android/jvm `apiCheck`s green with no dump changes

Not run locally: the emulator-backed `iosSimulatorArm64Test` tasks — no firebase CLI on this machine. Relying on CI for those.

PR #875, opened 2026-09-07
Add com.google.firebase source-compatibility layer (header stubs on Android/JVM)

nbransby:
Make the SDK usable as a drop-in replacement for the Firebase Android SDK: code written against `com.google.firebase.*` (including `com.google.android.gms.tasks.Task` and its listeners) compiles unchanged in common code for Android, iOS, JVM and JS. No import changes, no dependency swap on Android.

## Compatibility layer (firebase-app, firebase-installations)

- `com.google.android.gms.tasks` (`Task`, listeners, continuations, `TaskCompletionSource`), `kotlinx.coroutines.tasks.await` and `com.google.firebase` (`Firebase`, `FirebaseApp`, `FirebaseOptions`, `Timestamp`, exceptions, `Firebase.initialize`) / `com.google.firebase.installations` are declared as `expect`s in `commonMain`, mirroring the Android SDK's names and shapes (Task results, listener interfaces, builders, `@JvmStatic` accessors). Declarations that are identical on every platform (the listener interfaces, `FidListener`) are plain common code rather than expect/actual.
- Apple actuals wrap the iOS SDK, JS actuals the modular JS SDK; both share pure-Kotlin implementations of `Task`, `Task.await()` and `Timestamp` in `nonJvmMain`.
- **Android/JVM actuals are header stubs**: signature-identical declarations that are compiled against, verified with ASM to match the real classes on the compile classpath (same JVM descriptors and static-ness), dumped for the compatibility report, and then deleted from the compilation output (`utils.stripHeaderStubs` in buildSrc). Nothing from `com.google.*` or `kotlinx.coroutines.tasks` ships in the AAR/JAR, so consumers, D8 and test runtimes bind to the real Firebase Android SDK (`firebase-java-sdk` on the JVM) and `kotlinx-coroutines-play-services` exactly as before. Other libraries depending on Firebase, the Firebase Gradle plugins and Play Services keep working.
- The shipped exceptions are top-level functions. In `FirebaseInitializeKt`: `Firebase.initialize(context: Any?, ...)`, `Firebase.getApps(context: Any?)` and `Firebase.fromResource(context: Any?)`. The SDK's `Firebase.initialize` takes a `Context`, which common code cannot name, so the facade ships real `Any?` overloads that cast and delegate to the SDK; Android code passing a `Context` still binds to the SDK's more specific overloads. `FirebaseApp.getApps(Context)` and `FirebaseOptions.fromResource(Context)` are static members of Java classes and cannot be given an `Any?` overload, so they move onto `Firebase` under the same name (calling the real static through a package-private Java helper on Android, since a stub's companion members compile to `Companion` calls; `FIROptions.defaultOptions()` on Apple; JS has no default configuration). In `TimestampInstantKt`: a top-level `Timestamp(instant: kotlin.time.Instant)` factory and `Timestamp.toKotlinInstant()`, the multiplatform counterparts of the JVM-only `Timestamp(Instant)`/`Timestamp(Date)` constructors and `toInstant()`/`toDate()`; plain common code that also binds to the Android SDK's `Timestamp` (a top-level function rather than a companion extension because `firebase-java-sdk`'s `Timestamp` has no companion). `kotlin.time.ExperimentalTime` is opted in at the build level.
- Android-only members that have a multiplatform replacement (`FirebaseApp.initializeApp`, `getApps`, `getApplicationContext`, `FirebaseOptions.fromResource`, the `Date`/`Instant` constructor, `toDate` and `toInstant` of `Timestamp`) are declared with `@Deprecated(level = ERROR)` and the platform type widened to `Any`: common code using them fails to compile with a message naming the replacement, with a `ReplaceWith` quick-fix where the replacement is a drop-in expression (a `Timestamp(instant)` call with an `Instant` argument resolves to the new factory on its own), while Android code keeps binding to the real SDK member. The stub verifier skips these members since they cannot be called. Members with no replacement to point at (`Timestamp`'s `Parcelable` members) are omitted. `setAutomaticResourceManagementEnabled` is a no-op outside Android. APIs available on Android, JVM and Apple but not JS go in a `nonJsMain` source set (`utils.applyFirebaseHierarchy()`).
- The `dev.gitlive.firebase` wrappers of both modules now live in `commonMain` on top of the compat layer (exposed as `.compat`); the platform `actual`s are reduced to the `.android` / `.ios` / `.js` accessors. The exceptions are typealiases of the `com.google.firebase` ones.

## Verification (buildSrc `compat/`)

- `androidSourceCompatDump` / `androidSourceCompatCheck` compare `api/android/<module>.api` plus the header-stub dump with the vendored Android SDK `api/android-sdk/*.api.txt` (Metalava format, pinned to the `androidSdk.apiRef` commit, refreshed with `updateAndroidSdkApi`) and write `api/android-sdk-compat.txt`. Members provided through an accepted mapping, including the error-deprecated ones (reported as `MAP` with their message), count as available; `exclusions.txt` lists the intentional omissions: `@hide` and `@platform` ones are not counted, the rest count as unavailable, and `Class#member(Type, Type)` selects a single overload. Top-level functions are matched across file facades. The report's percentage is the module's README coverage badge: both firebase-app and firebase-installations are at 100%. `check` runs the check, the PR workflow runs the dump alongside `apiDump`.
- `AndroidSdkSourceCompatTest` exercises the layer from `commonTest` exactly as Android app code would (`Firebase.initialize` of a named app, `Firebase.getApps`, `Firebase.fromResource`, static accessors, Task listeners, `TaskCompletionSource`, `registerFidListener`, `Timestamp` and its `Instant` conversions) on every platform; `FirebaseInstallationsTest` covers the suspend wrappers. Installations tests are skipped on the JVM because `firebase-java-sdk` does not support Installations. iOS simulator tests link keychain entitlements (as firebase-auth does) since the Installations SDK stores the FID in the keychain.

## Notes

- Not a breaking change: the `.android` accessors still return the Google types, JVM facade names are preserved (`FirebaseKt`, `CommonKt`, `InstallationsKt`), the version stays 2.7.0. New public API: `FirebaseApp.compat`, `FirebaseInstallations.compat`, `com.google.firebase.FirebaseInitializeKt` (`Firebase.initialize(Any?, ...)`, `Firebase.getApps(Any?)`, `Firebase.fromResource(Any?)`), `com.google.firebase.TimestampInstantKt` (`Timestamp(Instant)`, `Timestamp.toKotlinInstant()`, exposing the still-experimental `kotlin.time.Instant`); on Android `kotlinx-coroutines-play-services` becomes an `api` dependency of firebase-app.
- Fixes a pre-existing JS bug: the Installations externals imported `delete` instead of the SDK's `deleteInstallations`.
- Rules for code inside the migrated modules: the stubs' companion members compile to `Companion` calls that do not exist at runtime, so static SDK members are reached through the SDK's Kotlin extensions (`Firebase.app`, `Firebase.installations`, `Firebase.initialize`) or a package-private Java helper.
- Docs: `documentation/migrate-from-android.md` is a step-by-step guide to moving Android SDK code into a shared module, including the compiler-guided replacement of the Android-only members (its before/after code was compiled as probes on JS, JVM and Android); README ("Using the Firebase Android SDK API from common code" and "Initialization") and `.github/copilot-instructions.md` document the two API layers and the rules for adding to them.


nbransby:
CI note on `7fa881b`: `build-js (cleanTest :firebase-database:jsTest)` failed in the browser run only, with `FirebaseDatabaseTest.testOrderByKeyWithStartAtAndEndAt` throwing `DatabaseException: Database not connected`. That is the test's `ensureDatabaseConnected()` pre-check timing out against the Realtime Database emulator before the test body runs; the node run of the same suite passed.

This PR does not change firebase-database (its diff against master is empty), and the same job passed on the two previous runs of this branch (`7dee0f8` and `0c6d275`). The last commit only touched buildSrc's ASM dependency and the JVM facade names in firebase-app / firebase-installations. I'll re-run the failed job once the workflow run completes.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #770, opened 2025-08-31
Tvos

Reedyuk:

PR #740, opened 2025-08-01
bumped kotlin and dependancies.

Reedyuk:

PR #894, opened 2026-09-23
Re-enable iOS tests for firebase-storage

nbransby:
Adds `ios` back to `firebase-storage.supportedTestTargets`, and fixes the test setup that was crashing the storage tests on iOS.

### Why

PR CI builds only what it tests. With no iOS test target for storage, no workflow compiled its Apple code: not the nightly, which uses the same filter, and not the lint/API job, which only covers jvm/android. A compile error in `appleMain` could merge without anyone noticing. #893 nearly shipped one; a separate cross-compile caught it.

History of the skip:
- **June 2024:** toggled off and on while chasing a CI hang ("temp skip storage tests on iOS to see if it explains hanging").
- **Aug 2025:** 03ff64d6 turned it off again in #740, the PR that got CI working again. No specific reason was recorded.
- **Oct 2025:** #770 replaced `skipIosTests` with `supportedTestTargets`, and the skip carried over.

### The actual failure

With the tests re-enabled, the first run failed with this:

```
Current test: testDownloadData
FirebaseStorage/Storage.swift:218: Fatal error: Cannot connect to emulator after Storage SDK initialization.
Call useEmulator(host:port:) before creating a Storage reference or trying to load data.
```

firebase-ios-sdk caches `Storage` instances by app name and bucket (`InstanceCache` in `FirebaseStorage/Sources/Storage.swift`) and never evicts them. The tests deleted the default app after each test and recreated it before the next one, so `Firebase.storage(app)` returned the previous test's instance. That instance had already been used, and calling `useEmulator` on it crashed the process. The crash kills the whole test run, which fits the old "hang" reports.

The fix follows the pattern the auth tests already use (0f9dd653): create the app once, apply `useEmulator` only when it's created, and stop deleting it between tests. Android and JS behave the same way, since `useEmulator` is only valid before first use and the instance is cached per app there too.

### Verified locally

These ran against a real Storage emulator (firebase-tools 15.30.2):
- JS: all 16 storage tests pass.
- wasmJs: all 18 pass, including `WasmStorageInteropTest`, which creates and deletes its own default app.

iOS can only be checked here in CI. `v3.0.0` is merged in, so this run also covers #893's `DataTest` on the iOS simulator. `gradle.properties` sits outside every module, so CI runs the full matrix, not just storage.


nbransby:
`build-wasm-js (cleanTest :firebase-database:wasmJsBrowserTest)` is failing here, but not because of this PR. It only runs because `gradle.properties` changed, which triggers the full matrix, and it already fails on `v3.0.0` itself. The last full run on `v3.0.0` (run 33645576410 at 769f7e1a, this PR's original base) failed on exactly this job, and it was the only failure in that run. The symptom is identical both times: Karma reports `Disconnected (0 times), because no message in 30000 ms` before any test result comes back.

No fix exists for it yet. It needs its own investigation in `firebase-database`, so it's out of scope here. I'll re-run the job once after this run finishes.

---
_Generated by [Claude Code](https://claude.ai/code)_

nbransby:
A second database job failed here too, again unrelated to this PR: `build-js (cleanTest :firebase-database:jsTest)`. `FirebaseDatabaseTest.testOrderByKeyWithStartAtAndEndAt` failed in the browser run only, with `DatabaseException: Database not connected`. It passed in the node run of the same job and in the last full run on `v3.0.0`.

That exception comes from `awaitWhileOnline` in `firebase-database/src/jsMain/.../database.kt`. It races every write against `.info/connected` and throws as soon as that reports `false`, so a brief connection drop in the browser fails the `setValue`. The fix is already on `master` as #878 (8164bb1, "Let Realtime Database writes suspend while offline instead of failing"), but it hasn't been merged into `v3.0.0` yet. `v3.0.0`'s `wasmJsMain` has its own copy of `awaitWhileOnline` that would need the same change. Porting it touches database on every platform, so it should go in its own PR rather than this one. I'll re-run the failed jobs once after this run finishes.

---
_Generated by [Claude Code](https://claude.ai/code)_

